PR #44, opened 2018-05-14
Simpifies the validation model

### LINKED ISSUE #40: Simplify validation model: eliminate invalid states
The state record:
```fs
type Model = 
    { TempF: double
      TempC: double
      Error: string option }
```
works fine for small pages and little components but when your page becomes large (for example when you have a form) having multiple error flags clutters the model and forces the developer to keep the fields values in-sync with their error flags. A better solution in my experience is to encode the value in a single explicit type:
```fs
type Temperature =
   | Value of double 
   | ValidationError of string

type State = 
  { TempF : Temperature
    TempC : Temperature } 
```
this way you can't possibly use the data from `Temperature` if there was a parse/validation error

Zaid-Ajaj:
Closes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,10 +158,10 @@ so they can be correctly and simply displayed to the user.  Here is an example o
         { TempF: Temperature
           TempC: Temperature }
 
-    /// Valdiate a temperature in fareneit, can be shared between cliemt/server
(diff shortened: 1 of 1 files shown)


dsyme:
Thanks!

Issue #48, opened 2018-05-16
Build fails with VS 15.7: cannot find provided types 


Zaid-Ajaj:
After working around #47 and restoring packages successfully, I am not able to build the solution from Visual Studio 2017 latest and I get errors from the fsharp compiler that look like related to type providers:

| Severity | Code  | Description | Project | File | Line | Suppression State |
| --------- | ------ | ------------  | --------- | --------- | ------- | -------|
| Error |FS0193 |Could not find file 'C:\projects\elmish-xamarin-forms\Samples\AllControls\Droid\ProvidedTypes5c2b5dd0-36e9-4067-97f5-fdfe2f421972.dll'.	 |AllControls.Droid |FSC	|1| Active|
|Error|	FS0193|	Could not find file 'C:\projects\elmish-xamarin-forms\Samples\TicTacToe\Droid\ProvidedTypes9f0bff89-67dd-46c2-a040-82a1e59c9ec5.dll'.	|TicTacToe.Droid|	FSC	|1	|Active|

dsyme:
A hotfix for this issue should be out in VS2017 15.7 soon.

dsyme:
Note Visual Studio 15.6.* doesn't have this problem, also VS for Mac is ok.  If you need a workaround you can also use the FSHarp.Compiler.Tools pacakge

dsyme:
Or you can target iOS

dsyme:
Or you can just play with the apps that don't use static resources (e.g. CounterApp)

nosami:
> I am not sure I understand what this means, reference the package from the projects with errors?


Try this - https://stackoverflow.com/a/43845651/316874

dsyme:
@nosami Is it possible to adjust the samples to remove the uses of the Android resource type provider for now? It is only used in fairly minor ways for now

dsyme:
Fixed in #51

PR #51, opened 2018-05-17
Remove Android resource type provider.

nosami:
This commit should be reverted once
https://github.com/Microsoft/visualfsharp/pull/4860
is fixed.

Re #48


dsyme:
Thanks!  I'll also open a new PR immediately to revert this since I think we eventually want it there.

PR #53, opened 2018-05-17
Fix #45 : Modified Generator to Reference Properties updates  NOT use UpdateIncremental when types of Old and New object differs

### LINKED ISSUE #45: Dynamically changing "ContentPage --> content" property is not updated correctly
My demo code is like below

```fsharp
type Msg = Next
type Model = { Count: int }
let init () = { Count = 0 }, Cmd.none
let update msg model =
    match msg with | Next -> { Count = if model.Count > 0 then 0 else 1 }, Cmd.none
let view model dispatch =
    Xaml.ContentPage(
        title = "Demo",
        content =
            if model.Count = 0 then
                Xaml.Button(text = "Next1", command = fun _ -> dispatch Next)
            else
                Xaml.StackLayout(
                    children = [
                        Xaml.Label(text = "Finished")
                        Xaml.Button(text = "Go back", command = fun _ -> dispatch Next) ]
                )
    )
```
When I click next the Count is change to 1 and program is also running in to else block, but the UI is not updated.

gibranrosa:


dsyme:
LGTM, surprised something in the samples didn't flush this out

dsyme:
0.4.4

PR #54, opened 2018-05-17
Attempt OSX apps build (Don't merge!)

nosami:
Attempting to build on OSX so that we can build sample apps.


dsyme:
It will be so good to have this, both for correctness and as a guide to others.

nosami:
I'm still looking into it, it's not at all obvious how to make it all work :)

nosami:
So, this is actually building now on OSX. Looks like some legitimate warnings and errors to fix up.

dsyme:
OK, I pushed the necessary build fixes and we have green!! This is fantastic, thanks @nosami I could never have worked that stuff out.

PR #57, opened 2018-05-19
Fix build errors with FsPickler.Json

dsyme:


dsyme:
This was included in #54

PR #58, opened 2018-05-20
Add WPF platform to the AllControls Sample

mydogisbox:
An attempt at adding WPF to the AllControls Sample as mentioned in https://github.com/fsprojects/Elmish.XamarinForms/issues/55 .


dsyme:
I could run the sample - I noticed a few glitches, which I assume are due to the Xamarin.Forms.WPF implementation.

The sample isn't very "pretty" and it would be great if someone went over the controls and made them look a bit better (e.g. made the editing box be a decent size, used StackLayout a bit less etc.) :)

mydogisbox:
Yeah I saw that.  There are definitely some re-sizing issues going on at a minimum.

Are you suggesting a view function just for WPF?

dsyme:
Will close/reopen to retest CI

nosami:
I hit the same error earlier because the log file has become too long. Cherry-picking https://github.com/fsprojects/Elmish.XamarinForms/pull/69/commits/ac24b4fa5a9b7ae2deba63b2f625c401d6dde6d9 should fix it.

Once that is fixed, I think the build will still fail because we don't have WPF on the OSX build machine :)

dsyme:
replaced by #175

PR #60, opened 2018-05-21
scrolling larger  lists on-demand

dsyme:
This does a little work on scrolling large lists.  

* the `itemAppearing` and `itemDisappearing` events now pass the index of the item appearing/disappearing
* AllControls contains a sample

We need to do more work on large list views, e.g. 
* give a `keyFunction` to get the key for an item
* don't use linear searches through the list of items


dsyme:
On my Google Pixel 2 the somewhat naive "infinite scrolling" sample continues to perform well even with 13,000 items in the ListView. (note, no backing data is stored, but the list of items is being re-specified frequently, though is only growing at the end).

There are a large number of optimizations we could perform from here but that's already pretty good.

PR #68, opened 2018-06-13
Add FormattedText parameter to Label. 

liboz:
Also adds a FormattedString creation method along with FormattedText and Spans methods for the XamlElement type


dsyme:
Hi @liboz!

This DynamicXaml.fs file is generated - could you try modifying `bindings.json` appropriately and re-run the code generation command at the top of the `Program.fs` file?

Also if you could add a "sanity test" by adding something that utilizes this to the AllControls sample that would be great.

thanks
don

liboz:
@dsyme Okay I have done as you suggested.

dsyme:
This is great!

PR #80, opened 2018-06-25
add template, SearchBar support, CI testing of template instantiation

dsyme:


dsyme:
There are still some glitches on the full testing on travis, but I will pull this since the nuget has been pushed

PR #83, opened 2018-06-26
try-block for each command

### LINKED ISSUE #82: Possible incomplete command dispatch
https://github.com/fsprojects/Elmish.XamarinForms/blob/e5048a7b84eca911ecdee332f08a74b0111cc3b0/Elmish.XamarinForms/ElmishProgram.fs#L66-L69

At the moment it is just a theoretical issue. On Elmish that is not a problem because if there is an error the original state is reverted, but in this case, the new model is set before the commands are dispatched. So having an exception on the first command can prevent the other commands to be dispatched risking having the model in between states that can't be updated.

Probably that would work ok:
```fsharp
for sub in newCommands do
    try 
        sub dispatch
    with ex ->
        program.onError ("Error executing commands:", ex)
```

As a plus, with that change I would be able to create a client for Elmish.Bridge using the `onError` to send server messages without needing to rewrite features of the parent library as I [needed](https://github.com/Nhowka/Elmish.Bridge/blob/master/src/Client/Library.fs#L133-L197) with Elmish. I couldn't use the `onError` because of the model reverting on error and that also made the model trapped between states on a test project.

Nhowka:
Prevents an incomplete command dispatch. Closes #82

### DIFF
--- a/Elmish.XamarinForms/ElmishProgram.fs
+++ b/Elmish.XamarinForms/ElmishProgram.fs
@@ -63,10 +63,11 @@ type ProgramRunner<'model, 'msg>(app: Application, program: Program<'model, 'msg
                 updateView updatedModel 
             with ex ->
                 program.onError ("Unable to update view:", ex)
-            try 
-                newCommands |> List.iter (fun sub -> sub dispatch)
-            with ex ->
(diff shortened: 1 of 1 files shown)


dsyme:
Thanks!

PR #93, opened 2018-07-02
use an attribute builder

dsyme:
The process of building attribute property bags was overly expensive, involving calls with 20-30 parameters even when most of them are `null`. In addition the attribute arrays were copied several times. 

This modifies to do one pass to count the attributes, one to fill them in.  

We additionally inline the creation methods so that most of this can be pre-computed based on the properties actually provided

Also document writing wrappers for 3rd party extensions


dsyme:
image

PR #97, opened 2018-07-03
Add support for SkiaSharp and more type safe extension API

dsyme:


dsyme:
This adds initial support for SkiaSharp

PR #110, opened 2018-07-11
Allow fscd to build on Mono

nosami:


dsyme:
Moved to net45 in #109

PR #134, opened 2018-07-22
Fixes to ListViewGrouped

TimLariviere:
All fixes only apply to ListViewGrouped
- Fixed tryFindGroupedListViewItem (wrongly assumed item would be ListGroupData instead of ListElementData)
- Fixed itemAppearing and itemDisappearing (when grouped, the 2 events are also called for the group headers)
- Hardcoded IsGroupingEnabled=true and removed it from the bindings (Setting it to false defeats the purpose of the control)


dsyme:
Lovely, thanks!

PR #138, opened 2018-07-24
Added an embedded image into the AllControls sample

TimLariviere:


Samples/AllControls/AllControls/AllControls.fs
```diff
@@ -417,10 +417,14 @@ module App =
...
                dependsOn () (fun model () -> 
                  View.NonScrollingContentPage("Image", 
-                     [ View.Label(text="Image:")
+                     [ View.Label(text="Image (URL):")
                        View.Image(source="http://upload.wikimedia.org/wikipedia/commons/thumb/f/fc/Papio_anubis_%28Serengeti%2C_2009%29.jpg/200px-Papio_anubis_%28Serengeti%2C_2009%29.jpg", 
                            horizontalOptions=LayoutOptions.FillAndExpand,
-                           verticalOptions=LayoutOptions.FillAndExpand) ]))
+                           verticalOptions=LayoutOptions.FillAndExpand)
+                       View.Label(text="Image (Embedded):", margin=Thickness(0., 20., 0., 0.))
+                       View.Image(source=ImageSource.FromResource("AllControls.Baboon_Serengeti.jpg", typeof<RootPageKind>.Assembly),
```
dsyme: What happens if you just use the name of the resource here as a string?  I presume it doesn't work?

TimLariviere: No, it doesn't work.
It would assume that the string is a path to an image file (usually from resources in native projects).
To load an embedded image, we have to specify the assembly which contains it.

PR #140, opened 2018-07-25
Extensions - Fixed namespaces by extending the correct View class

TimLariviere:
The extensions open both `Elmish.XamarinForms.DynamicViews` and `Xamarin.Forms`.
As both have a `View` class, the latest namespace opened was selected for extending with `type View with`.
This resulted in a wrong namespace for `View.Map` and `View.SKCanvasView`.

```fsharp
open Xamarin.Forms
View.Map()
```
Instead of
```fsharp
open Elmish.XamarinForms.DynamicViews
View.Map()
```

I fixed it by fully qualifying the `View` class.
I could have just inverted the open statements, but figured it would be asking for trouble later.

Made it clear in the docs as well.


dsyme:
Ah yes, thanks! :)

PR #143, opened 2018-07-27
Added support for Jump List in ListViewGrouped

TimLariviere:
Xamarin.Forms ListView supports Jump List by setting the property
`GroupShortNameBinding = new Binding("KeyName")`

This is mostly an iOS specific thing, as Android just ignores this. (Maybe Windows uses it?)

I did not exposed the property itself in the bindings json file, as it would be awkward to create a binding in full elmish and people wouldn't know what key name to put in.

Instead I just expose a boolean property `showJumpList` which will set the binding property if needed.
`ListGroupData` has a new string member `ShortName` which stores the name of the group to display in the jump list.
This property is defined when setting `ListViewGrouped.items`. It now requires a `string * ViewElement * ViewElement[]` signature for each group.

What I don't like with my approach is that it requires to define the `ShortName` of a group, even if we don't want a Jump List (which might be most of the times)

Implementation is up for debate.

ios-jumplist

### DIFF
--- a/Elmish.XamarinForms/ViewConverters.fs
+++ b/Elmish.XamarinForms/ViewConverters.fs
@@ -23,10 +23,11 @@ type ListElementData<'T>(key:'T) =
     member x.Key = key
 
 [<AllowNullLiteral>]
-type ListGroupData<'T>(key:'T, coll: 'T[]) = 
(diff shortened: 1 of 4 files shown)


TimLariviere:
I've had confirmed that his property is also used on UWP.

dsyme:
It looks good.  I think you've made the right tradeoffs

PR #147, opened 2018-07-28
Added DisplayAlert and DisplayActionSheet

TimLariviere:
Added simple helpers around methods `Page.DisplayAlert` and `Page.DisplayActionSheet`.
Those are statically accessible through `Application.Current.MainPage`, just like the TicTacToe sample does.

But accessing the MainPage is confusing when doing full-elmish, as we don't have direct access to UI elements most of the times.

### DIFF
--- a/Elmish.XamarinForms/ViewHelpers.fs
+++ b/Elmish.XamarinForms/ViewHelpers.fs
@@ -172,3 +172,12 @@ module SimplerHelpers =
                 externalsTable.Add(externalObj, res)
                 res
 
+    type View with
+        static member displayAlert title message cancel =
+            Application.Current.MainPage.DisplayAlert(title, message, cancel) |> Async.AwaitTask
+
+        static member displayAlertWithConfirm title message accept cancel =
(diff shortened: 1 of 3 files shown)


dsyme:
Hmmmm..... :)

* They would be PascalCase (since we're generally following F# naming conventions so far)
* They would be better uncurried I think (callers can then use named argumentsm, and it's a case where named arguments makes the calling code clearer)
* We would need to duplicate the XML docs for these

But when you add all that up it doesn't quite feel worth it.  And they can be defined in user code (e.g. docs or a sample)

TimLariviere:
Indeed.
I agree that uncurried would be better.

Ok, so I will remove the helpers and write some docs on how to use dialogs in full-elmish.
I'll revert the TicTacToe sample, but keep what I did on the AllControls sample.

dsyme:
@TimLariviere Super, thanks

TimLariviere:
I close this PR. I will open a new one when I'm done.

PR #150, opened 2018-08-01
Fixed NavigationPage Reuse

### LINKED ISSUE #148: Swapping page type inside NavigationPage does not work
When using a NavigationPage, swapping the root ContentPage by another one works and UI is refreshed. (No navigation)
But when swapping to another type (like TabbedPage), UI is not refreshed.
However without a NavigationPage, everything is fine.

Here's a reproduction of the problem:
https://github.com/TimLariviere/EXF_ReproNavPageBug

TimLariviere:
Fixes #148

I've opted to add a special case in the function `canReuseChild` by checking if at least one page has changed type.

Not sure if the code is great though. (conversion to Seq to avoid exception with Zip when the 2 arrays have a different length)

### DIFF
--- a/Elmish.XamarinForms/ViewConverters.fs
+++ b/Elmish.XamarinForms/ViewConverters.fs
@@ -135,27 +135,24 @@ module Converters =
 
     let identical (x: 'T) (y:'T) = System.Object.ReferenceEquals(x, y)
 
-    // NavigationPage can be reused only if the already existing pages don't change type (added/removed pages don't prevent reuse)
-    // E.g. If the first page switch from ContentPage to TabbedPage, the NavigationPage can't be reused; otherwise it can.
-    let canReuseNavigationPage (prevChild:ViewElement) (newChild:ViewElement) =
-        let prevPages = prevChild.TryGetAttribute<ViewElement[]>("Pages")
-        let newPages = newChild.TryGetAttribute<ViewElement[]>("Pages")
-
-        match prevPages, newPages with
-        | ValueSome prevPages, ValueSome newPages ->
-            let prevSeq = Seq.ofArray prevPages
-            let newSeq = Seq.ofArray newPages
-            Seq.forall2 (fun (prev: ViewElement) (curr: ViewElement) -> prev.TargetType = curr.TargetType) prevSeq newSeq
-        | _, _ -> true
-
-    let canReuseChild (prevChild:ViewElement) (newChild:ViewElement) =
+    let rec canReuseChild (prevChild:ViewElement) (newChild:ViewElement) =
         if prevChild.TargetType = newChild.TargetType then
-            if newChild.TargetType = typeof<NavigationPage> then
+            if newChild.TargetType.IsAssignableFrom(typeof<NavigationPage>) then
                 canReuseNavigationPage prevChild newChild
             else
                 true
         else
             false
+
+    // NavigationPage can be reused only if the pages don't change their type (added/removed pages don't prevent reuse)
+    // E.g. If the first page switch from ContentPage to TabbedPage, the NavigationPage can't be reused.
+    and canReuseNavigationPage (prevChild:ViewElement) (newChild:ViewElement) =
+        let prevPages = prevChild.TryGetAttribute<ViewElement[]>("Pages")
+        let newPages = newChild.TryGetAttribute<ViewElement[]>("Pages")
+
+        match prevPages, newPages with
+        | ValueSome prevPages, ValueSome newPages -> (prevPages, newPages) ||> Seq.forall2 canReuseChild
+        | _, _ -> true
 
     let updateChild (prevChild:ViewElement) (newChild:ViewElement) targetChild = 
         newChild.UpdateIncremental(prevChild, targetChild)
--- a/Elmish.XamarinForms/ViewConverters.fs
+++ b/Elmish.XamarinForms/ViewConverters.fs
@@ -135,8 +135,26 @@ module Converters =
 
     let identical (x: 'T) (y:'T) = System.Object.ReferenceEquals(x, y)
 
-    let canReuseChild (prevChild:ViewElement) (newChild:ViewElement) = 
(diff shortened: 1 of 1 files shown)


Elmish.XamarinForms/ViewConverters.fs
```diff
@@ -135,8 +135,26 @@ module Converters =
...
-        (prevChild.TargetType = newChild.TargetType)
+    // NavigationPage can be reused only if the already existing pages don't change type (added/removed pages don't enter into account)
+    // E.g. If the first page switch from ContentPage to TabbedPage, the NavigationPage can't be reused; otherwise it can.
+    let canReuseNavigationPage (prevChild:ViewElement) (newChild:ViewElement) =
+        let prevPages = prevChild.TryGetAttribute<ViewElement[]>("Pages")
+        let newPages = newChild.TryGetAttribute<ViewElement[]>("Pages")
+
+        match prevPages, newPages with
+        | ValueSome prevPages, ValueSome newPages ->
+            Seq.zip (prevPages |> Seq.ofArray) (newPages |> Seq.ofArray)
```
dsyme: Looks like this could be simplified to `Array.forall2`?

dsyme:
> Not sure if the code is great though. (conversion to Seq to avoid exception with Zip when the 2 arrays have a different length)

Oh just saw this comment - check the lengths are equal then call Array.forall2?  The lengths do need to be equal to allow reuse, right?

TimLariviere:
>The lengths do need to be equal to allow reuse, right?

No. Otherwise we would create a new NavigationPage each time we're pushing/popping pages.
We only want to check the pages that are still here.

`Seq.forall2` seems to do the job. Thanks!
I didn't know that one existed. :)

Elmish.XamarinForms/ViewConverters.fs
```diff
@@ -135,8 +135,27 @@ module Converters =
...
+    // NavigationPage can be reused only if the already existing pages don't change type (added/removed pages don't prevent reuse)
+    // E.g. If the first page switch from ContentPage to TabbedPage, the NavigationPage can't be reused; otherwise it can.
+    let canReuseNavigationPage (prevChild:ViewElement) (newChild:ViewElement) =
+        let prevPages = prevChild.TryGetAttribute<ViewElement[]>("Pages")
+        let newPages = newChild.TryGetAttribute<ViewElement[]>("Pages")
+
+        match prevPages, newPages with
+        | ValueSome prevPages, ValueSome newPages ->
+            let prevSeq = Seq.ofArray prevPages
+            let newSeq = Seq.ofArray newPages
```
dsyme: Array is automatically compatible with Seq, so you can remove these two conversions, thanks!

Also this is cleaner:

            (prevPages, newPages) ||> Seq.forall2 (fun prev curr -> prev.TargetType = curr.TargetType) 

In fact should this be

            (prevPages, newPages) ||> Seq.forall2 canReuseChild

just in case we add further special cases to `canReuseChild` for other types?

Elmish.XamarinForms/ViewConverters.fs
```diff
@@ -135,8 +135,27 @@ module Converters =
...
+            Seq.forall2 (fun (prev: ViewElement) (curr: ViewElement) -> prev.TargetType = curr.TargetType) prevSeq newSeq
+        | _, _ -> true
+
+    let canReuseChild (prevChild:ViewElement) (newChild:ViewElement) =
+        if prevChild.TargetType = newChild.TargetType then
+            if newChild.TargetType = typeof<NavigationPage> then
```
dsyme: I think this should use `IsAssignableFrom` (a subtyping test) rather than strict type equality, just in case someone defines a subclass of `NavigationPage`

Elmish.XamarinForms/ViewConverters.fs
TimLariviere: Good point. Without `||>`, forall2 required explicit typing, so that's solved.

For canReuseChild, do you mean I should use a mutually recursive function to call canReuseChild inside canReuseNavigationPage?

Elmish.XamarinForms/ViewConverters.fs
TimLariviere: Ok

dsyme:
@TimLariviere I think I should make you a package owner as I'll be going on vacation for a couple of weeks.  Would that be ok?  What's your nuget id? thanks

TimLariviere:
@dsyme Ok, my nuget id is TimotheLariviere.

Maybe I should take you up on your offer to be co-admin as well?
That way, I can merge small bugfixes while you're on vacation.
Thanks

dsyme:
@TimLariviere Notes on building and releasing are here; https://github.com/fsprojects/Elmish.XamarinForms/blob/master/DEVGUIDE.md#dev-notes---releasing

Bumping the version number is a PITA since it occurs in several places.  Just search-and-replace down the whole directory tree using VSCode, be sure not to miss anything.  

You should bump the minor version number if there has been any change in API.  If in doubt just bump it.

TimLariviere:
@dsyme Ok. All packages should be kept in sync?

dsyme:
@TimLariviere Yes, all packages published in sync, including templates.  Note the build process instantiates the template  - if recycling builds using the same version number it's a good idea to do a regular

    git clean -xfd .

and also  clean out Elmish.XamarinForms from your `.nuget` directory, otherwise I think the template build may end up using previous packages with the same version number. So once you get used to it you can be generous in regularly bumping package versions :)

dsyme:
@TimLariviere Feel free to push simplifications to make this, well, simpler...

TimLariviere:
@dsyme From what I see, I only need to set the version in the RELEASE_NOTES.md, LiveUpdate.fs (help text) and template.json?
The other files are AssemblyInfo.fs that are generated by FAKE.

Is that all the places where we need to change the version?

dsyme:
@TimLariviere Yes - though I always just search down the entire code tree.  Would be great to eventually derive the version only from RELEASE_NOTES.md :)

PR #178, opened 2018-09-08
Upgrade to Xamarin.Forms 3.1

### LINKED ISSUE #145: Upgrade to XF 3.1
I was going to add a missing property [ListView.SelectionMode](https://docs.microsoft.com/dotnet/api/xamarin.forms.listview.selectionmode), but found this was new in XF 3.1 ([link](https://github.com/xamarin/Xamarin.Forms/issues/1801)).

So my question is:
Can we update EXF to use the latest version of XF (templates and all), or is it tied to XF 3.0 for a reason?

TimLariviere:
Xamarin.Forms 3.1 has been stable for 3 months now, so I think we can update.

I've also updated the bindings with the new properties from https://developer.xamarin.com/releases/xamarin-forms/xamarin-forms-3.1/3.1.0/#whats-new-in-this-release

(PS: Paket does not seem to manage the packages of the .NET Standard projects, only the Android/iOS projects. Wouldn't it be better to make Paket handle everything?)

Closes #145

### DIFF
--- a/Fabulous.Core/Fabulous.Core.fsproj
+++ b/Fabulous.Core/Fabulous.Core.fsproj
@@ -21,6 +21,6 @@
   <ItemGroup>
     <PackageReference Include="FSharp.Core" Version="4.5.2" />
     <PackageReference Update="FSharp.Core" Version="4.5.2" /> <!-- workaround for VSMac bug https://github.com/mono/monodevelop/pull/5137 --> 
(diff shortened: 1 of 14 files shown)


dsyme:
> (PS: Paket does not seem to manage the packages of the .NET Standard projects, only the Android/iOS projects. Wouldn't it be better to make Paket handle everything?)

Yes it would :)

Issue #185, opened 2018-09-17
Provide a way to get at the underlying Xamarin control


kevinha:
Please. Use case is showing the DatePicker dialog from a different control such as a button or icon. This is done by calling `Focus()` on the Xamarin DatePicker control. I don't think a 'focus' attribute is applicable here as it would require Fabulous to track the state of the dialog. Access to the Xamarin control may also be useful for animation up the until the proper solution for animation has been implemented.

Zaid-Ajaj:
This is similar to [Refs in React](https://reactjs.org/docs/refs-and-the-dom.html) very important mechanism for interop with the original DOM element: allows for animations, third-party library support and custom incremental-update logic (among other uses)

dsyme:
Yes, we need this for #183 and #182

kevinha:
Also would make it easier to work around some control shortcomings, e.g. TimePicker lack of 'TimeChanged' callback could be solved with PropertyChanged callback without recourse to a rendering.

TimLariviere:
I completely agree that we need this.
But I also think we should emphasize on discouraging its use, in the docs.

For example, using it to avoid making a CustomRenderer is a bad practice for me.
This feels really hacky.

dsyme:
@TimLariviere Agreed. I looked at the react docs and they do a pretty good job of striking the right balance.

dsyme:
@TimLariviere There are several designs we could use for this.

One would be a `created` event handler for every control.

Another would be to try to do something more html/react-like and use some kind of `ref` property (a string) which is a handle usable with a `getref` function that can look up the latest value of the control in the overall view.  Xamarin Forms already have `AutomationId` - see https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.element.automationid?view=xamarin-forms and perhaps that should be used, renamed to "id".

Or we could do both

TimLariviere:
I like the idea of a `ref` handle.
This would be easier to use in most cases compared to the `created` callback.
It would also prevent common mistakes
- Accessible only in the local scope, no need for global mutable fields to store the value. (memory leaks)
- `tryGetRef` to prevent access to an element that has been destroyed, or not yet created.

dsyme:
> •	Accessible only in the local scope, no need for global mutable fields to store the value. (memory leaks)

I think the Refs would typically have long lifetime.  But they could hold weak handles to the items, so no leaks

dsyme:
[...]

Issue #183, opened 2018-09-15
Supporting Animations


willsam100:
The roadmap in this repo states that supporting animations is coming. 
Being able to use the XamarinForms animations would be a huge win for this project. 

Is there an estimation/guess on when this work will be started/completed. 

Alternatively, is there planned architecture for this, allowing the community to contribute?

dsyme:
It's a great question. The relevant C# version of the functionality is here: https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/animation/simple.  Basically each view element can now become a task in a little animation DSL. There are also added features of easing functions and custom animations.

This raises a number of issues: views are non-temporal and thus animations are a problem.  I'd like to know what Elmish frameworks do for this, short of programming up the animation manually.

I assume views would need to include a limited amount of temporal information, but obviously you can't restart animations on every view invocation. There is also the question of cancelling animations, and running them in parallel (including parallel animations of unrelated items).

For example, perhaps ViewElements could describe their future animation but subject to a model-held token which indicates the `start` of the animation, e.g.

```
let update msg model = 
    match msg with 
    ... -> { model with Animate= StartAnimation() }

let view ... = 
    ....
    View.Image(...).WithAnimation(model.Animate, ScaleTo (1, 2000))
    ...
```
Cancellation would then be `model.Animate.Cancel()` at some point.

This however doesn't cover animation of two otherwise unrelated view elements, e.g. 
```
await Task.WhenAny<bool>
(
  image.RotateTo (360, 4000),
  image.ScaleTo (2, 2000)
);
```
Maybe you could hack in something like this:
```
let image1 = None
let image2 = None

let update msg model = 
    match msg with 
    | ... -> { model with Animate= StartAnimation() }

let view ... = 
    View.StackLayout [
       View.Image(..., created= (fun v -> image1 <- Some v));
       View.Image(..., created= (fun v -> image2 <- Some v))
    ]..WithAnimation(model.Animate, 
              Animation.WaitAll [
                     Animation.ScaleTo (image1,1, 2000); 
                     Animation.ScaleTo (image2,1, 2000)
              ])
    ...
```
but it's not exactly ideal.  And async/task programming is actually a lovely way of specifying animations and it would be kind of good to use it.

dsyme:
Hi @willsam100 

Animations appear to be possible with #190 

[...]

Issue #182, opened 2018-09-13
SkiaSharp Canvas does not redraw


GuuD:
**Applies to:** Android, possibly iOS
**Description:**
Regardless of model updates, SkiaSharp Canvas never redraws, and stuck on first drawn image.

I suspect the issue is not calling .InvalidateSurface() method on underlying SKCanvasView, but I don't know enough about Xamarin approaches to be sure that this is idiomatic way of triggering repaints. 

If this is indeed a root cause, should I file pull request that adds .InvalidateSurface() call to update function of SkiaSharp.fs?
```fsharp
let update (prevOpt: ViewElement voption) (source: ViewElement) (target: SKCanvasView) = 
    View.UpdateView (prevOpt, source, target)
    //Add this to trigger redraw on all but first invocations of update
    match prevOpt with
    | ValueSome _ -> target.InvalidateSurface()
    | _ -> ()
    source.UpdatePrimitive(prevOpt, target, CanvasEnableTouchEventsAttribKey, (fun target v -> target.EnableTouchEvents <- v))
    source.UpdatePrimitive(prevOpt, target, IgnorePixelScalingAttribKey, (fun target v -> target.IgnorePixelScaling <- v))
    source.UpdateEvent(prevOpt, PaintSurfaceAttribKey, target.PaintSurface)
    source.UpdateEvent(prevOpt, TouchAttribKey, target.Touch)
```

TimLariviere:
Quickly skimming through SkiaSharp's docs, and seems you're right. MSDN recommends using `InvalidateSurface` to force a redraw -- even though any other `Invalidate*` method should be used really carefully in Xamarin/WPF for performance reasons.

I don't think invalidating the surface at each update is a good idea.
The view gets redrawn for any message sent by your app, and doing so might add flickering on your canvas.

**Example**: You have a page with an Entry and a SKCanvasView. To track what the user inputs, you listen to `textChanged` and send a message each time.
This means that the canvas will get redrawn at each keystroke, even though you don't need to update it.

Maybe we should expose a new parameter that provides with an "Invalidate" action when the control is created?
```fsharp
static member SKCanvasView(?paintSurface: (SKPaintSurfaceEventArgs -> unit),
                           ?touch: (SKTouchEventArgs -> unit),
                           ?enableTouchEvents: bool,
                           ?ignorePixelScaling: bool,
                           ?loaded: ((unit -> unit) -> unit))

/// App

[...]

PR #190, opened 2018-09-19
SkiaSharp invalidate and add "created" event

dsyme:
Possible resolutions for  #182 and #185 and also helping with #183 

See discusssion in #185 

@TimLariviere let me know what you think.  We may need to adjust this further, it's not set in stone


Fabulous.Core/ViewElement.fs
```diff
@@ -45,6 +46,9 @@ type ViewElement internal (targetType: Type, create: (unit -> obj), update: (Vie
     static member Create (create: (unit -> 'T), update: (ViewElement voption -> ViewElement -> 'T -> unit), attribsBuilder: AttributesBuilder) =
         ViewElement(typeof<'T>, (create >> box), (fun prev curr target -> update prev curr (unbox target)), attribsBuilder.Close())
 
+    [<System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)>]
+    static member val _CreatedAttribKey : AttributeKey<obj -> unit> = AttributeKey<_>("ElementCreated")
```
TimLariviere: Is it feasible to expose the real type through `created: ('T -> unit)`?
This would help avoid bad conversion, and expose the hidden custom types (CustomListView).

dsyme:
Note `created` is not strongly typed when it could be. I actually think we should change that, we need to adjust the code generator for that

dsyme:
> Is it feasible to expose the real type through created: ('T -> unit)? This would help avoid bad conversion, and expose the hidden custom types (CustomListView).

Yes, I think we should specialize it to `self` as it goes down the inheritance chain.

TimLariviere:
This seems good.

If I'm not mistaking, we would use it like that?
```fsharp
let mutable myEntry : Entry option = None

let update msg model =
     match msg with
     | FocusEntry ->
              match myEntry with
              | Some e -> e.Focus()
              | None -> ()
              model

let view model dispatch =
      View.ContentPage(
            content=View.StackLayout(
                   children=[
                           View.Button(title="Focus Entry", command=(fun() -> dispatch FocusEntry))
                           View.Entry(created=(fun e -> myEntry <- Some (e :?> Entry)))
                   ]
             )
       )
```

The more I think about this, the more I think we need to be clear in the docs about the implications.
This could potentially lead to memory leak (not clearing the mutable fields when leaving the page).

dsyme:
> If I'm not mistaking, we would use it like that?

Yes.  Though the focus on mutable state is not great - the "refs"/"id" approach seems more strongly typed:

```
let myEntry = ViewReference<Entry>()

let update msg model =
     match msg with
     | FocusEntry ->
              match myEntry with
              | Some e -> myEntry.Value.Focus()
              | None -> ()
              model

let view model dispatch =
      View.ContentPage(
            content=View.StackLayout(
                   children=[
                           View.Button(title="Focus Entry", command=(fun() -> dispatch FocusEntry))
                           View.Entry(ref=myEntry)
                   ]
             )
       )
```

> The more I think about this, the more I think we need to be clear in the docs about the implications. 

yes

TimLariviere:
Ah yes, way better.

When would `myEntry.Value` be resolved?
Might be good to cache the control in a WeakRef table.

dsyme:
@TimLariviere this is just about ready but needs trialling. No tests or docs yet, and really needs to be trialled with animations, focus etc

It likely needs tweaking for cases where the element changes or has not yet been created.

dsyme:
I've implemented `ViewRef` and made `created` strongly typed.  

Example for animation triggered on creation (note: untested as yet due to technical issue on my laptop)

    View.Label(text="Rotate", created=(fun l -> l.RotateTo (360.0, 2000u) |> ignore)) 

Example for animation triggered on message (note: untested as yet due to technical issue on my laptop)

    let animatedLabelRef = ViewRef<Label>()

    let update msg model =
        match msg with 
        | Poked -> animatedLabelRef.Value.RotateTo (360.0, 2000u) |> ignore

    let view dispatch model = 
        ...
           View.Label(text="Rotate", ref=animatedLabelRef) 
           View.Button(text="Rotate", command=(fun () -> dispatch Poked)) 
        ....
```

dsyme:
This is complete but we just need to test an app runs ok on device (CI testing in this repo doesn't actually test execution of much Fabulous code as yet, just a lot of compilation of it).

dsyme:
I've tested this on device and the animations added in AllControls work.

I also made the liveupdate work for multiple file projects.

I'll do some more testing of live update

Issue #184, opened 2018-09-17
Fabulous - Drop Paket?


TimLariviere:
Paket is used awkwardly in Fabulous. It doesn't really manage anything.

It is used to download FAKE 4, download some dependencies and lock version for the NuGet packages (paket.template).
But the .NET Standard libs manage their own dependencies using PackageReference, and the samples only directly reference the downloaded dependencies (Paket doesn't manage the samples).
https://github.com/fsprojects/Fabulous/blob/58bbfd79558f964270eeb66ba75acddea93f50da/Samples/AllControls/Droid/AllControls.Droid.fsproj#L91

It's quite confusing.

I would like to uniformize the repo by going either full Paket or full NuGet.
Problem is:
- Paket is not supported by Visual Studio for Mac (breaks IntelliSense, Restore working only with `msbuild /t:Restore`)
- Paket doesn't seem to support [dotnet/SourceLink](https://github.com/dotnet/sourcelink) (something I'd like to add later)
- We have to make sure by hand that Paket and Fabulous use the same version of the dependencies

So I wonder: Maybe it would be better to drop Paket and go full NuGet (PackageReference)?
We can achieve the same "one place to manage all the dependencies" with a `Directory.Build.props` file.

What do you think?

dsyme:
Yes, in this repo I've only really use paket to get a package resolution that I trust and then do the rest by hand.  

The problem is that it's sort of a sanity check and a transitive package resolution I trust. I trust paket much more than `dotnet` to tell me about inconsistencies between packages (too often `dotnet` seems to give warnings rather than errors).

I'm not sure what to do about it.  For this repo I don't want to use paket more than we do, and if we could continue to relegate it's role that is fine.  But I don't want to give up on the sanity checking.  Tricky

dsyme:
> We can achieve the same "one place to manage all the dependencies" with a Directory.Build.props file.

I think this approach may be consistent with using paket as a sanity checker.  Basically the dependencies listed in Directory.Build.props would be a resolution of (some of) the transitive dependencies, and we can check the versions from a lock file against the props?

TimLariviere:
I'm currently working with `Directory.Build.props` to find a solution to this issue.
The solution I came up with is to fix package versions in a `props` file.
```xml
<Project>
    <!-- Package versions -->
    <PropertyGroup>
        <FSharp_Core_Version>4.5.2</FSharp_Core_Version>
[...]

PR #195, opened 2018-09-26
Centralized packages versionning with Directory.Build.props / FAKE 5 / SourceLink

TimLariviere:
This took a while, but I'm finally pushing it to Fabulous. :)

This PR tackles partially #184 

Several enhancements regarding the tooling for Fabulous contributors.
- Centralized packaging metadata (authors, version, notes) into a single place : Directory.Build.props
- Mostly centralized packages version into a single place : Packages.targets
- Removed all `paket.template` files; Uses `dotnet pack` instead
- Removed all AssemblyInfo.fs files (those are generated automatically by `dotnet build`)
- Added support for SourceLink (allow to step into Fabulous source code right from the nupkgs)
- Migrated build script to FAKE 5
- <s>Updated build machines to iOS 12 and Android 9</s> (Travis XCode 10 is not compatible with Xamarin.Android)

There's still 2 places where the packages are managed: Paket and Packages.targets
The next step should be to generate Packages.targets from the versions resolved by Paket, but I don't know if this is possible.

Concerning SourceLink, it's a "Pack" only dependency, it has no use when building.
This would be fine normally, but it's not compatible with `msbuild` (which is used by the samples)...
<s>So to prevent that, I had to do something a little dirty in the FAKE script:
```
dotnet add package SourceLink
dotnet pack
dotnet remove package SourceLink
```
I ended up doing this after lots of unsuccessful tries, because VS for Mac has a terrible support for MSBuild Condition...</s>

EDIT: While trying to make a repro to file an issue on MonoDevelop, I made the SourceLink.props file work with VS for Mac. So this is way better now,

@slang25 @nosami let me know what you think


TimLariviere:
Travis is currently failing when building the samples because it can't find the Android SDK.
```
error XA5300: The Android SDK Directory could not be found. Please set via /p:AndroidSdkDirectory. [/Users/travis/build/TimLariviere/Fabulous/samples/AllControls/Droid/AllControls.Droid.fsproj]
```
But I don't really know why. :/
Even if I pass the value through `/p:AndroidSdkDirectory` it doesn't work.

nosami:
> I ended up doing this after lots of unsuccessful tries, because VS for Mac has a terrible support for MSBuild Condition...

Please could you file a bug report for this at https://github.com/mono/monodevelop ? VS for Mac uses MSBuild directly these days to evaluate conditions, so a bit puzzling.

nosami:
I _think_ the build is failing because Travis has Java 9 installed and the Android SDK is dependent on Java 8

nosami:
Maybe try installing Java 8 (suggested by homebrew)

```
android-sdk requires Java 8. You can install it with
  brew cask install homebrew/cask-versions/java8
```

or try this to make it work with Java 9 https://stackoverflow.com/a/47346176/316874. I guess this would involve running `sed` on `sdkmanager` to replace the line that is suggested.

nosami:
Looks like adding this to the start of script https://github.com/fsprojects/Fabulous/blob/master/install-xamarin-android-ios.sh might work too.

`export JAVA_OPTS='"-Dcom.android.sdklib.toolsdir=$APP_HOME" --add-modules java.xml.bind' would work too (instead of trying to modify the `sdkmanager` script.

nosami:
🍺  android-sdk was successfully installed!

TimLariviere:
```Please could you file a bug report for this at https://github.com/mono/monodevelop ? VS for Mac uses MSBuild directly these days to evaluate conditions, so a bit puzzling.```
Yes, as soon as I have repros, I'll open some issues.

**Comment by dsyme** [comment]
Wow, fantastic work!

**Comment by nosami** [comment]
```
ResolveSdks: /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(665,2): error XA0030: Building with JDK Version `10.0.2` is not supported. Please install JDK version `1.8.0`. See https://aka.ms/xamarin/jdk9-errors
```

:(

https://docs.microsoft.com/en-us/xamarin/android/troubleshooting/questions/jdk9-errors Looks like we need to install OpenJDK 1.8.0.9 (it's what we're using internally at the moment, we should be switching from the Oracle distribution soon)

I can get the build working on a separate PR if you want - it will need to wait until tonight though.

**Comment by nosami** [comment]
Need to script the guidance on this page https://docs.microsoft.com/en-us/xamarin/android/get-started/installation/openjdk

File: .travis.yml [hunk 1]
```diff
@@ -1,7 +1,7 @@
 language: csharp
 
 os: osx
-osx_image: xcode9.3
+osx_image: xcode10
```
nosami: Was this change needed?

It looks like it changes the version of Java that is installed

1.8.0 https://travis-ci.org/fsprojects/Fabulous/builds/433555917#L70

vs

10.0.2 https://travis-ci.org/fsprojects/Fabulous/builds/433529887#L59

TimLariviere: I don't think so.
Changed it as part of the global update (along with Xamarin.iOS and Xamarin.Android)
But it can be reverted if needed.

nosami: Looks I came to the same conclusion independently [from you](https://github.com/fsprojects/Fabulous/pull/196#issuecomment-424747145) :)

Looks like reverting this change is the simplest solution providing everything else works.

nosami:
```
MTouch: MTOUCH(0,0): error MT0091: This version of Xamarin.iOS requires the iOS 12.0 SDK (shipped with Xcode 10.0). Either upgrade Xcode to get the required header files or set the managed linker behaviour to Link Framework SDKs Only (to try to avoid the new APIs).
```

I just reopened my PR with Java 8 and the XCode 10 Travis image. Let's see what happens.

TimLariviere:
Finally fixed Travis by reverting to XCode 9.3 / Xamarin.iOS 11.
Thanks @nosami!

If you find a way to upgrade to XCode 10, that would be great. :)

dsyme:
I think this needs .NET CLI 2.1.401, I don't yet have that on my laptop
```
C:\GitHub\dsyme\Fabulous\samples\CounterApp\CounterApp>dotnet --version
2.1.202
```

dsyme:
Sorry about the conflicts

TimLariviere:
>I think this needs .NET CLI 2.1.401, I don't yet have that on my laptop
Hmm, interesting. When rewriting the build script, I had force the version of `dotnet`, but this is no longer the case.
Would it be FAKE 5 DotNet module that force a specific version?

TimLariviere:
NuGet API is down... Waiting before retrying build

TimLariviere:
@dsyme Ok, found why you have trouble with `dotnet`
>Note that .NET Core SDK 2.1.300 or newer is required for SourceLink to work.
https://github.com/dotnet/sourcelink#prerequisites

I should add a "Prerequisites" paragraph in the DEVGUIDE

PR #200, opened 2018-10-01
fix type of ViewRef TryValue

### LINKED ISSUE #199: ViewRef<'T>.TryValue returns obj option rather than 'T option
Or the documenation at https://fsprojects.github.io/Fabulous/views-animations.html#animations is wrong:

```
match animatedLabelRef.TryValue with 
| None -> () 
| Some c -> c.RotateTo (360.0, 2000u) |> ignore
```

dsyme:
Fixes #199


dsyme:
@TimLariviere If you feel the need to crank the handle on the next release then this can be included as part of it :)

PR #201, opened 2018-10-01
ListView row reusing

### LINKED ISSUE #198: ListView does not reuse it's 'items' ViewElements
In the attached repro the items of the ListView are labels and the text of the first label changes in response to a scheduled message. According to debug logs the labels are recreated every time the view function is called and none are reused. This appears to be due to the canReuse function at ViewConverters.fs#L271

`updateCollectionGeneric (ValueOption.map seqToArray prevCollOpt) (ValueOption.map seqToArray collOpt) targetColl ListElementData (fun _ _ _ -> ()) (fun _ _ -> false) (fun _ _ _ -> failwith "no element reuse")`

always returning false.

In the repro there is a reference to my local debug build of Fabulous.Core which will need to be replaced.

TimLariviere:
Fixes #198 

`IListElement` now implements `INotifyPropertyChanged` to allow the ViewElementCell to do an incremental update whenever the row changes, instead of destroying it and creating a new one.
See https://github.com/fsprojects/Fabulous/issues/198#issuecomment-425716261 for more information

### DIFF
--- a/src/Fabulous.Core/ViewConverters.fs
+++ b/src/Fabulous.Core/ViewConverters.fs
@@ -323,7 +323,7 @@ module Converters =
                 let oc = ObservableCollection<ListGroupData<'T>>()
                 target.ItemsSource <- oc
                 oc
-        updateCollectionGeneric prevCollOpt collOpt targetColl ListGroupData (fun _ _ _ -> ()) (fun _ _ -> false) (fun _ _ _ -> failwith "no element reuse")
+        updateCollectionGeneric prevCollOpt collOpt targetColl ListGroupData (fun _ _ _ -> ()) (fun (_, prevKey, _) (_, currKey, _) -> canReuseChild prevKey currKey) (fun _ (_, currKey, _) target -> target.Key <- currKey)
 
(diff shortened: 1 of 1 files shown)


dsyme:
This is fantastic, great work

Can I ask what validation you did?  e.g. in AllControls?

Also, we should consider how we would add unit tests for this sort of thing.  It doesn't have to be in this PR, just saying that I left a testing hole, focusing on samples more than unit tests

TimLariviere:
I validated it against the repro zip from #198 and checked that AllControls was still working.
It could use more tests, especially regression tests.

AllControls is quite hard to go into, and not a good place for testing as this is supposed to be a sample.
Maybe it would be better to start a new project in the `tests` folder focused on integration tests?

dsyme:
Yes we need a new test project, eg also for direct testing of ViewElement etc. Would have found #199 before merge to master

dsyme:
BTW AllControls isn't really a sample now, I'm happy to keep extending  it as a dumping ground for adhoc scenario tests that we can actually run that explore stuff that can't easily be pinned down by new unit tests.

Issue #205, opened 2018-10-09
Missing documented property of StyleClass on VisualElement


davidglassborow:
The [styles](/fsprojects/Fabulous/blob/master/docs/views-styling.md#css-styling-with-xamarinforms-30) documentation uses a styleClass property that doesn't seem to exist on any of the view objects.

I'm keen to do a PR to add it, so:

- The underlying property is on [VisualElement](https://docs.microsoft.com/en-gb/dotnet/api/xamarin.forms.visualelement.styleclass?view=xamarin-forms) so I'm guessing we need a new entry in Xamarin.Forms.Core.Json under `Xamarin.Forms.VisualElement`
- The underlying control is an `IList<string>` but the docs suggest a simple string, so should I be splitting the F# string or similar when applying to the underlying control ?
- How do I run the generator ?  It doesn't seem to run as any of the fake targets, and running from the CLI or VS Mac leads to a `Unhandled Exception: System.IO.FileNotFoundException: Could not load file or assembly 'FSharp.Core, Version=4.5.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.` error

TimLariviere:
Hi @davidglassborow 

Thanks for your interest in Fabulous. :)
`styleClass` is indeed missing.

Based on your remarks, I've fixed a few things with the generator. (see https://github.com/fsprojects/Fabulous/pull/206 - pending merge)
- No more missing FSharp.Core
- FAKE target to run the generator (docs here : https://github.com/fsprojects/Fabulous/pull/206/commits/b868fe8fd0118a68ac3bee05ef26d60d66ac7f89)

>The underlying property is on VisualElement so I'm guessing we need a new entry in Xamarin.Forms.Core.Json under Xamarin.Forms.VisualElement

Yes, exactly.

>The underlying control is an IList<string> but the docs suggest a simple string, so should I be splitting the F# string or similar when applying to the underlying control ?

I think splitting the string would not be the best option.
It would be better to use a list of string, but what the docs of Fabulous suggest is also a valid option (easier to write).
eg.
```fsharp
View.Label(text = "Hello", styleClass = "myStyle")
View.Label(text = "Hello", styleClass = [  "myStyleA"; "myStyleB" ])
```

You can take a look at how Margin is implemented (supports both string and Thickness)
[...]

PR #209, opened 2018-10-10
Adding StyleClass to VisualElement

davidglassborow:
see #205

### DIFF
--- a/extensions/Maps/Xamarin.Forms.Maps.fs
+++ b/extensions/Maps/Xamarin.Forms.Maps.fs
@@ -29,7 +29,7 @@ module MapsExtension =
                                  ?horizontalOptions, ?verticalOptions, ?margin, ?gestureRecognizers, ?anchorX, ?anchorY, ?backgroundColor,
                                  ?heightRequest, ?inputTransparent, ?isEnabled, ?isVisible, ?minimumHeightRequest, ?minimumWidthRequest,
                                  ?opacity, ?rotation, ?rotationX, ?rotationY, ?scale, ?style, ?translationX, ?translationY, ?widthRequest,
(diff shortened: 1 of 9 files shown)


TimLariviere:
Well done!
Do you know if there's some limitation with changing/removing css?

I tried some cases with the AllControls app, but it doesn't seem to update correctly.
- "cssCallout" changed to "cssRed" (.cssRed { color: Red; })
- "cssCallout" changed to ""

They don't update visually, even though the control itself is properly updated

davidglassborow:
Maybe an XF bug ?  Looking at https://github.com/xamarin/Xamarin.Forms/issues/3824, maybe related.  Same issue seems to exist with StyleId.

CSS is pretty new in XF, so maybe we can assume a XF limitation for the moment ?

Doesn't affect my use case for the moment so I'm keen to get this merged and nuget updated if your happy ?

dsyme:
The `styleClass` parameter should be added for each of the three extensions, e.g. here https://github.com/fsprojects/Fabulous/blob/master/extensions/Maps/Xamarin.Forms.Maps.fs#L32

It's annoying to have to do this but not too hard.

dsyme:
Looks like the Unix build is failing, I'm not totally sure why

TimLariviere:
Yes, noticed it as well.
It fails for a different reason each time (aapt missing, missing NuGet package, etc.), but most of the times it works.

davidglassborow:
CI / Paket failing for some reason.  That package is available on nuget so no idea why travis is failing.

```
Paket failed with
-> Could not download OpenTK.GLControl 3.0.1.
-> Couldn't get package details for package OpenTK.GLControl 3.0.1 on https://api.nuget.org/v3/index.json.
```

PR #226, opened 2018-11-12
Generator - Add support for types with base type in another assembly

TimLariviere:
Add a cache of loaded assemblies in the generator to support types with one or more base types in other assemblies.
This allows for scenario where we want to use a custom type that extends the base Xamarin.Forms types.

```json
"assemblies": [
    "packages/neutral/Xamarin.Forms/lib/netstandard1.0/Xamarin.Forms.Core.dll",
    "src/Fabulous.Core/bin/Release/netstandard2.0/Fabulous.Core.dll",
],
```

```json
    {
      // Extends and replaces Xamarin.Forms.Entry
      "name": "Fabulous.DynamicViews.CustomEntry",
      "modelName": "Entry",
      "members": [
        {
          "name": "PropertyThatOnlyExistsInCustomEntry",
          "defaultValue": "null"
        },
        {
          "name": "Text",
          "defaultValue": "null"
        },
```

Currently if Mono.Cecil tries to load the base types of a custom control (Fabulous.Core.dll) that are found in another assembly (Xamarin.Forms.Core.dll), Mono.Cecil throws an exception because it can't find Xamarin.Forms.Core

The need comes from this issue: https://github.com/fsprojects/Fabulous/issues/225

### DIFF
--- a/tools/Generator/Program.fs
+++ b/tools/Generator/Program.fs
@@ -97,6 +97,30 @@ type TextWriter with
     member this.printf fmt = fprintf this fmt
     member this.printfn fmt = fprintfn this fmt
 
+type RegistrableResolver() =
+    inherit BaseAssemblyResolver()
+
+    let cache = Dictionary<string, AssemblyDefinition>()
+
+    override this.Resolve(name) =
+        match cache.TryGetValue(name.FullName) with
+        | true, assembly -> assembly
+        | false, _ ->
+            let assembly = base.Resolve(name)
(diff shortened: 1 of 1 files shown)


garethhubball:
Would it be worth me grabbing this fork for tonight's stream? I'm streaming in just over an hour and working with this would be really fun.

TimLariviere:
@garethhubball Yes, this would be a good way to validate these changes against a real example. :)

Issue #225, opened 2018-11-06
EntryCell has no TextChanged Event


garethhubball:
I'm trying to use an EntryCell in my application and the `completed` event only fires when the done button is pressed on the keyboard. This is Xamarin.Forms intended behaviour.

There is, however, no simple way that I can find of getting an event to fire when the focus is lost or when the text has changed. This makes the EntryCell far less useful in my opinion. Could we add an event to it somehow in Fabulous or find some other way of getting the text from said EntryCell without resorting to methods that make the architecture impure?

garethhubball:
Would anyone mind if I had a go at implementing this? I don't know the source well so it might be a big ask for the support but I think it might be nice to contribute to the project. I'll also add a request to the Xamarin.Forms repo and have a crack at implementing it there too.

TimLariviere:
You're welcome to implement it. :)
Don't hesitate to ask if you need help.

As a starting point, you can take a look at ListView and its custom control.
The idea is to create a custom control that will support `TextChanged`
https://github.com/fsprojects/Fabulous/blob/e6407ce0058152f123d9faba1f559906b9af7d95/src/Fabulous.Core/ViewConverters.fs#L109-L110

Then you can use your control in the bindings file and add your event.
https://github.com/fsprojects/Fabulous/blob/e6407ce0058152f123d9faba1f559906b9af7d95/tools/Generator/Xamarin.Forms.Core.json#L1592-L1593

TimLariviere:
Yes, the Generator tries to infer some infos like the type of the property/event by doing reflection.
But it uses the field "Name" (currently set to `Xamarin.Forms.Entry`) to find the control, so it won't find your new event, thus the error.

So to do that, you'll need to do it a bit differently than the CustomListView.

First you'll need to load `Fabulous.Core.dll` at the start of the json file
```json
"assemblies": [
    "packages/neutral/Xamarin.Forms/lib/netstandard1.0/Xamarin.Forms.Core.dll",
    "src/Fabulous.Core/bin/Release/netstandard2.0/Fabulous.Core.dll",
],
```
Then you'll need to set your new control in the `Name` field.
But we still want the Generator to use the name "EntryCell". You can force it by setting the `ModelName` field
And now you can add your new property:
```json
    {
      // Extends Xamarin.Forms.EntryCell
      "name": "Fabulous.DynamicViews.CustomEntryCell",
      "modelName": "EntryCell",
      "members": [
        {
          "name": "TextChanged",
[...]

PR #227, opened 2018-11-13
Add a TextChanged Event to the EntryCell

garethhubball:
Started work on the `TextChanged` event to start fixing #225 - it seems to work in WPF but not in iOS/Android. The `Text` property seems to be ignored in iOS/Android and needs looking into.

Please don't merge this yet as it does not work!


TimLariviere:
PR #226 needs to be merge before this one.
Also a rebase will be needed once done.

TimLariviere:
One last thing :)
You need to add a reference to `Fabulous.CustomControls.dll` in the template projects and samples projects as well, like you did for `AllControls`

TimLariviere:
I think I know why this doesn't work on Android and iOS.
You're hiding the base property `Text` of EntryCell with the one you declared in CustomEntryCell.
Android and iOS still refer to `EntryCell.Text`, and without override they won't call your code.

For some reason, WPF is OK with that though.

garethhubball:
Thank you - I'll have a look at this tonight as well as cleaning up the PR.

garethhubball:
> I think I know why this doesn't work on Android and iOS.
> You're hiding the base property `Text` of EntryCell with the one you declared in CustomEntryCell.
> Android and iOS still refer to `EntryCell.Text`, and without override they won't call your code.
> 
> For some reason, WPF is OK with that though.

Ahhh, Jsut had a look at Xamarin.Forms and the base `Text` property isn't virtual, so while we can hide it, we can't simply override it. I wonder if I need to have a closer look at trying to do this in Xamarin.Forms directly and just bubble up the change once it's in there.

TimLariviere:
Another way you could do it is by subscribing to the PropertyChanged event of EntryCell yourself.
And if the property changed is `Text`, then you can trigger your TextChanged event.

garethhubball:
Why did I not think of that? Thank you!!!

garethhubball:
This now works across all three platforms. If there's anything I can do to make any merge easier please let me know - I still haven't really got the hang of PRs, Git nor working in open source yet so apologies if and when I go wrong.

TimLariviere:
Build is failing.
Could you add a reference to Fabulous.CustomControls in the fsproj of the templates?
https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/NewApp.Android/NewApp.Android.fsproj
https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/NewApp.WPF/NewApp.WPF.fsproj
https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/NewApp.iOS/NewApp.iOS.fsproj
https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/NewApp.macOS/NewApp.macOS.fsproj

If you want to validate these changes without waiting for Travis CI, you can run `.\build.cmd Test` on your machine

TimLariviere:
And finally, it would be great if you could move Fabulous.CustomControls into the `src` folder.

No need to exclude `Fabulous.CustomControls` from the `Build` step. This step will run `dotnet pack` on all src/**/*.fsproj, which we'll need to push the new project in NuGet.
You can just change your `BuildCustomControls` to do a `MSBuild Release` instead (only generate the dll for the Generator). Make sure you have the correct path to the dll in the bindings json afterwards.

TimLariviere:
This looks good. Well done!
I'll do a little preparation work and then will be able to release this in the next version :)

PR #229, opened 2018-11-14
Fix update for ListViewGrouped

TimLariviere:
Fixed recycling update for ListViewGrouped.
ListViewGrouped does not currently create new rows for an already existing group when recycling.
Before #201, it worked because all the rows/groups were recreated each time.


dsyme:
Lovely work!

PR #240, opened 2018-11-30
Dispatch function accessible in App projects

### LINKED ISSUE #239: Add easy way to dispatch from app projects
It would be nice if there was an easy way to dispatch messages from the App projects.

For example permissions - these are requested asynchronously in the Android and iOS app projects, and once accepted or rejected a message could then be dispatched to update the model and UI.

TimLariviere:
Closes #239 

The `dispatch` function is exposed in the `ProgramRunner` instance created when launching the application.
It can be later called to dispatch a message in Fabulous.

### DIFF
--- a/docs/update.md
+++ b/docs/update.md
@@ -169,3 +169,76 @@ let update msg model =
     | TakePicture -> model, (Cmd.ofAsyncMsgOption takePictureCmd)
     | PictureTaken -> ...
 ```
+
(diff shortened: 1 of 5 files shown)


TimLariviere:
Will add docs later.
Do not merge now.

TimLariviere:
Docs added.

Issue #243, opened 2018-12-02
Update to Xamarin.Forms 3.4


TimLariviere:
We're currently using Xamarin.Forms 3.1.
XF 3.3 will be 2 months old in a few days, so it might be good to update and provide wrappers for the new controls.

rbrogan-git:
Any reason not to go to XF 3.4?

TimLariviere:
We're waiting 2 months after a release before moving to it as a general rule, to wait for the first few bug fixes.
But I think we can safely go to XF 3.4 now.

rbrogan-git:
@TimLariviere I submitted a pull request for this one.

TimLariviere:
Closed by #257

PR #257, opened 2018-12-15
Update to XF 3.4 complete.

rbrogan-git:
Issue #243 
I am not an expert in paket but I think this is all that is needed, correct?  I ran the builds and tests after this update successfully.  I also launched the tic tack toe and all controls samples for android and all is working fine.

Let me know if this is good or of you need anything else.  

Thanks!
-Roger


slang25:
There's also the `paket.lock` file which should be updated too. You can do it by running `paket install`, that will update it and reflect the changes you've made to the dependencies file.

rbrogan-git:
@slang25 Thanks that is what I wanted to know.  Added the file.

slang25:
Looks like this also needs to change: https://github.com/fsprojects/Fabulous/blob/master/tests/Fabulous.Cli.Tests/Tests.fs#L14

rbrogan-git:
Thanks!  Found a another place with hard coded string versions.  Feels like a code smell to me.  Any reason we can pull this from paket?

PureWeen:
I think the wrappers also need to be updated with this

Here's an API diff. Not too many on existing controls and then there's a new ImageButton control
https://www.fuget.org/packages/Xamarin.Forms/3.4.0.1008975/lib/netstandard2.0/diff/3.1.0.697729/

rbrogan-git:
@PureWeen Thanks!  I has thinking some of that might show up as a compiler error.  I will look at the wrappers.

TimLariviere:
@rbrogan-git Sorry, should have been more explicit in the issue. Just bumping the version number is not enough. Fabulous requires a minimum version of Xamarin Forms (currently 3.1) but devs are free to update on their own.

What really interests us here is, like @PureWeen said, to update https://github.com/fsprojects/Fabulous/blob/master/tools/Generator/Xamarin.Forms.Core.json with the wrappers around the new controls, the new properties, the changed features, etc. between Xamarin.Forms 3.1 and 3.4.

You can take a look at a previous PR which updated from 3.0 to 3.1 : https://github.com/fsprojects/Fabulous/pull/178

slang25:
Could this be a 2 step process? Update the dependencies fixing any breaking changes, then PRs to lift through new changes. Or does that risk missing some of the apis?

I presume the json was semi-automated and now a manual task to add to?

TimLariviere:
The json file is fully manual for now (which causes some issues like forgotten controls or properties).
If we can find a way to automatically generate the json file and then customize it by hand, that would be awesome. (And open a way for 3rd parties)

I don’t feel like XF 3.4 is a major enough update to require a 2 step process.

Like you said, it would be hard to track every new features if we do it in separate PRs.

Also people might expect full support if we require them to update to XF 3.4.

TimLariviere:
@rbrogan-git Unfortunately no, we don't have that kind of documentation...
I'm currently refactoring the Generator, so it's a good time to write it.

The idea here is to add/update/remove any control/property/event accordingly in the `Xamarin.Forms.Core.json` file that have changed between XF 3.1 and 3.4.
This file is read at build time by the Generator which will generate the `Xamarin.Forms.Core.fs` file.
`Xamarin.Forms.Core.fs` is quite big and hard to read, you don't need to go into it.

(Tip: You can run the generator quickly with the command `.\build.cmd RunGenerator`)

The way the JSON file is defined is as such:
```
{
  "assemblies": [
    // Paths to dlls that will be loaded to do some reflection on it
    "packages/neutral/Xamarin.Forms/lib/netstandard1.0/Xamarin.Forms.Core.dll",
    (...)
  ],
	"outputNamespace": "Fabulous.DynamicViews",

        // List of controls for which to generate a Fabulous wrapper
	"types": [
		{
                        // Name of the actual implementation
			"name": "Xamarin.Forms.Element",

                        // List of properties or events of that control that we want to add to the wrapper
			"members": [ 
                                
				{
					"name": "ClassId",        // Name of the property
					"defaultValue": "null"    // Default value
                                },
                                (...)
                        ]
                },
               (...)
          ]
}
```
`name` and `defaultValue` are the strict minimum you need to add. The generator will try to infer everything else regarding `ClassId` using Reflection in the class `Xamarin.Forms.Element`. (Is it a property, an event? What type is it, bool, string? etc.)
If the property doesn't exist, the Generator will throw an error (non-existing properties are available, but its an advanced concept).

To enable more customization, there's a bunch of other attributes.
```
"name": "ItemsSource",
"defaultValue": "[]",

// Useful when there's multiple controls that have the same property name but are not related (different types)
"uniqueName": "ListViewGrouped_ItemsSource",

// Short name that will be used instead of the "name" in lower Pascal Case
"shortName": "items",

// What we expect people to provide in the View.MyControl(myProperty = value)
"inputType": "(string * ViewElement * ViewElement list) list",

// How we actually store the value (for performance reasons. Especially useful when we ask a list, but actually store it has an array)
"modelType": "(string * ViewElement * ViewElement[])[]",

// How we convert from "inputType" to "modelType"
"convToModel": "(fun es -> es |> Array.ofList |> Array.map (fun (g, e, l) -> (g, e, Array.ofList l)))",

// Custom update logic if the simple check between old and new value is not enough
"updateCode": "updateListViewGroupedItems"
```
`updateCode` refers to a function found here: https://github.com/fsprojects/Fabulous/blob/master/src/Fabulous.Core/ViewConverters.fs



**If we take an example:**
From the link of @PureWeen, we can see that there's a new `CornerRadius` property on `BoxView`, so we can go to the declaration of `BoxView` in the json file and add a new entry:
```json
{
	"name": "Xamarin.Forms.BoxView",
	"members": [
		{
			"name": "Color",
			"defaultValue": "Xamarin.Forms.Color.Default"
		},
		{
			"name": "CornerRadius",
			"defaultValue": "Unchecked.defaultof<Xamarin.Forms.CornerRadius>"
		}
	]
},
```
I used `Unchecked.defaultof` in `defaultValue` because CornerRadius is a struct (no null) and doesn't have a default value if we check the implementation in Xamarin.Forms
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/CornerElement.cs

**Another example:**
`WebView` has a new event `ReloadRequested`, so we can add it in the json file
```json
{
	"name": "Xamarin.Forms.WebView",
	"members": [
		{
			"name": "Source",
			"uniqueName": "WebSource",
			"defaultValue": "null"
		},
		{
			"name": "Navigated",
			"defaultValue": "null",
			"inputType": "Xamarin.Forms.WebNavigatedEventArgs -> unit",
			"convToModel": "(fun f -> System.EventHandler<Xamarin.Forms.WebNavigatedEventArgs>(fun _sender args -> f args))"
		},
		{
			"name": "Navigating",
			"defaultValue": "null",
			"inputType": "Xamarin.Forms.WebNavigatingEventArgs -> unit",
			"convToModel": "(fun f -> System.EventHandler<Xamarin.Forms.WebNavigatingEventArgs>(fun _sender args -> f args))"
		},
		{
			"name": "ReloadRequested",
			"defaultValue": "null",
			"inputType": "System.EventArgs -> unit",
			"convToModel": "(fun f -> System.EventHandler(fun _sender args -> f args))"
		}
	]
},
```

I hope it's clear.

rbrogan-git:
@TimLariviere This is my first pass on fixing up the json file.  I tried to add the TitleView property but was not successful so I kept it out for now. I am unclear on how to setup the convert function for View.

Suggestions welcome. Let me know if I missed anything.
Thanks!

TimLariviere:
@rbrogan-git Thanks! I'll take a look soon.
GitHub has some troubles to correctly diff the json file, due to the change of formatting.
To help with that, I've added a new build step that forces a specific formatting (independent of the IDEs).

Could you rebase your PR on the last commit on `master` please?
This will simplify the reviewing of your pull request. :)

rbrogan-git:
@TimLariviere I wondered why the diff got so weird.  Updated.  Thanks!

TimLariviere:
For TitleView, I still need to find the correct way to do it.
I'll keep you posted.

TimLariviere:
Oops, sorry for the conflict.

TimLariviere:
@rbrogan-git Here's a way to implement `TitleView`.

In the json file, we declare the property like usual but Xamarin.Forms is expecting a `Xamarin.Forms.View` and we only have a `ViewElement`.
So we will implement a custom update logic to convert our ViewElement to View (and implement incremental update too).
```json
{
  "name": "TitleView",
  "defaultValue": "null",
  "inputType": "ViewElement",
  "updateCode": "updateNavigationPageTitleView"
},
```

We then need to add this update function into https://github.com/fsprojects/Fabulous/blob/master/src/Fabulous.Core/ViewConverters.fs
There, we will take the previous ViewElement (if any), the new ViewElement (if any) and the NavigationPage to update.
Through the use of some existing functions, we can also implement incremental update.
```fsharp
let updateNavigationPageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Xamarin.Forms.NavigationPage) =
    match prevOpt, currOpt with
    | ValueSome prev, ValueSome curr when identical prev curr -> ()
    | ValueSome prev, ValueSome curr when canReuseChild prev curr ->
        updateChild prev curr (NavigationPage.GetTitleView(target))
    | _, ValueSome curr ->
        NavigationPage.SetTitleView(target, (curr.Create() :?> Xamarin.Forms.View))
    | ValueSome _, ValueNone ->
        NavigationPage.SetTitleView(target, null)
    | _, _ -> ()
```

_Might be some errors in this code. I haven't tested it yet_

rbrogan-git:
@TimLariviere thanks for all the information and help with this pull request.  I learned a lot about this project.  Sorry you ended up finishing it yourself.  I hope the work I did was helpful.

PR #262, opened 2018-12-22
SKCanvasView Always trigger redraw when invalidate is true

### LINKED ISSUE #261: SkiaSharp "invalidate" flag does not always trigger a redraw
Take this example code:

    let view dispatch model =
        dependsOn model.myValue (fun model myValue ->
            View.SKCanvasView(invalidate = true, paintSurface = ...)
        )

You would expect this code to just always redraw the canvas whenever the dependsOn causes a recompute, but this is not the case. ``invalidate`` is (wrongly) treated just like any other parameter, so it will only cause a redraw when set from ``false`` to ``true``.

jwosty:
Fixes #261


TimLariviere:
Thanks. Tested it against various cases (with and without dependsOn) and it looks good!

PR #274, opened 2019-01-08
Add mainpagebutton to each view

### LINKED ISSUE #137: AllControls sample - Add consistent back button for all pages
It would be great to have a consistent back button on all the pages of the AllControls sample, because currently it is quite painful to browse all the controls in the sample.
Most pages require to shutdown the app to go back to the RootPage.

Examples:
- `TabbedPage #1` has a back button hidden in the `Grid` tab
- `TabbedPage #3` has no back button at all

SergejDK:
Added a button "Main Page" to each view.

In the Flexlayout and Relativelayout example I needed to change the layout.

fix #137

### DIFF
--- a/samples/AllControls/AllControls/AllControls.fs
+++ b/samples/AllControls/AllControls/AllControls.fs
@@ -422,7 +422,6 @@ module App =
                                          View.BoxView(color).GridRow(i-1).GridColumn(j-1) ] )
                          MainPageButton
                        ]))
-
            ])
 
         | Tabbed1 ->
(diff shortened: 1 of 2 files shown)


TimLariviere:
Thanks! 👍 
That's great, we can now navigate into the app without being trapped. :)

Just one thing, the "Main Page" button is hard to reach in the pages with ListView, due to the double vertical scrolling.
Moving the button to the top for these two pages we would good.

Pages:
- TabbedPage # 2 > ListView
- Infinite scrolling ListView (especially this one as there's no other way to go back)

Issue #281, opened 2019-01-16
LiveUpdate failures


TimLariviere:
Here is a list of some cases where LiveUpdate fails to run.

- Assigning an event callback produces an "AddToObservable not supported" error (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-457137755)
- After a few iterations of simple changes the app started throwing OutOfMemoryException (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-455768379)
- Named fields in DU cases produces "couldn't find field NumberOfRepetitions in type DNamedType" (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-460947730)

Fixed:
- Using ViewRef<T> produces a "No member found for key Fabulous.DynamicViews.ViewRef" error (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-454885771) - Fixed by #286 
- Using a list a "No member found for key Microsoft.FSharp.Collections.FSharpList" error (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-455767701) - Fixed by #286 
- Calling a function that takes unit from a module in a separate dll produces a "No member found for key" error (https://github.com/fsprojects/Fabulous/issues/281#issuecomment-455538037) - Fixed by #286

TimLariviere:
From kevkov on Slack

Code:
```
let theRef = ViewRef<Label>()
```

Result:
```
{"Quacked":"couldn't quack! the evaluation of the decalarations in the code package failed: System.Exception: No member found for key (REntity Fabulous.DynamicViews.ViewRef`1[T], \".ctor\",\n RTypes [|Microsoft.FSharp.Core.Unit|])\n 
```

kevinha:
Assigning an event callback produces an "AddToObservable not supported" error, e.g.
`fun (entry:Entry) -> entry.Unfocused.Add(fun args ...`

kevinha:
Calling a function that takes `unit` from a module in a separate dll:

`let allSubjects = Database.getProfilesSummary ()`

produces

`android.runtime.JavaProxyThrowable: System.Exception: No member found for key (REntity Storage+Database, "getProfilesSummary", RTypes [|Microsoft.FSharp.Core.Unit|])`

while 

`let allSubjects = Database.getProfilesSummary 1`

is ok.

dsyme:
I'll try to get these fixed today, thanks

kevinha:
[...]

PR #286, opened 2019-01-19
fix interpreter bugs and make it pass tests

dsyme:
Fix most bugs in #281 

@TimLariviere It seems like a few of the Cli.Tests weren't passing, and so I patched things up - odd how many were failing.  Are they under CI?  I checked `build.fsx` and it looks like they are


TimLariviere:
Yes, they're under CI. But only Travis CI (macOS) actually runs them (FAKE target "Test").

Do you know which ones were failing?
Just tried to run them on my machine (macOS) and a VM (Windows), all are passing.

dsyme:
> Do you know which ones were failing? Just tried to run them on my machine (macOS) and a VM (Windows), all are passing.

Ah ok, it might be that my first fix caused others to fail, and so on - I didn't test them clean.  Now all are passing and the new tests are well, so all are good

TimLariviere:
@dsyme Is this PR ready to be merged?

dsyme:
@TimLariviere Yes it is, thanks

Issue #288, opened 2019-01-21
nuget.targets bug in Templates


jiangzhen3s:
https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/.nuget/NuGet.targets#L53

> <PaddedSolutionDir Condition=" '$(OS)' == 'Windows_NT'">"$(SolutionDir) "</PaddedSolutionDir>

==》    "$(SolutionDir)\\"

mallibone:
Hey @jiangzhen3s thanks for finding this one, I just stumbled over it myself. There seems to be another issue with that line that it ends with a `\` which is a problem when running `build.cmd Test` on Windows. I hope I'll have a fix later today so we can run the build and tests on Windows.

@TimLariviere this would mean we could also adapt [Azure DevOps CI](https://github.com/fsprojects/Fabulous/pull/285) to run the tests on every commit/pull request.

PR #290, opened 2019-01-21
Fix nuget errors when building/generating files from blank on Windows

mallibone:
Fixes errors when creating Fabulous from scratch on Windows and then using the dotnet template to generate projects. This also allows running `build.cmd Test` on a CI build.

Fix of issue #288


TimLariviere:
Thanks.

Closes #288

PR #293, opened 2019-01-23
#292 Move titileView from NavigationPage to Page.

### LINKED ISSUE #292: TitleView on NavigationPage doesn't work
The ```titleView``` argument of ```View.NavigationPage``` has no effect; nor does using the ```TitleView``` method on ```ViewElement```.

This is because ```NavigationPage.TitlePage``` is actually an attached property that needs to be applied to the pages you add to a ```NavigationPage```, rather than a property you set on ```NavigationPage``` itself.

chrisseed:
Fixes #292 by moving the TitleView property from ```NavigationPage``` to ```Page```.

Hopefully I've put it in the right place - Xamarin.Forms.Core.json looked to be sorted alphabetically with properties first, then events.

I've built and tested the nupkgs locally though I'm afraid I haven't updated the AllControls sample to add a usage of titleView/TitleView.

(This is my first (ever!) PR so please forgive me if I've done something wrong)

### DIFF
--- a/src/Fabulous.Core/ViewConverters.fs
+++ b/src/Fabulous.Core/ViewConverters.fs
@@ -627,7 +627,7 @@ module Converters =
         | ValueNone, ValueSome curr -> updateFunc defaultValue curr
 
     /// Update the attached NavigationPage.TitleView property of a Page, given previous and current values
-    let internal updateNavigationPageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Page) =
+    let internal updatePageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Page) =
         match prevOpt, currOpt with
         | ValueSome prev, ValueSome curr when identical prev curr -> ()
         | ValueSome prev, ValueSome curr when canReuseChild prev curr ->
--- a/src/Fabulous.Core/ViewConverters.fs
+++ b/src/Fabulous.Core/ViewConverters.fs
@@ -626,8 +626,8 @@ module Converters =
         | ValueSome prev, ValueNone -> updateFunc prev defaultValue
         | ValueNone, ValueSome curr -> updateFunc defaultValue curr
 
-    /// Update the TitleView of a NavigationPage, given previous and current values
-    let internal updateNavigationPageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Xamarin.Forms.NavigationPage) =
+    /// Update the attached NavigationPage.TitleView property of a Page, given previous and current values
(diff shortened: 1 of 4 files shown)


src/Fabulous.Core/ViewConverters.fs
```diff
@@ -626,8 +626,8 @@ module Converters =
         | ValueSome prev, ValueNone -> updateFunc prev defaultValue
         | ValueNone, ValueSome curr -> updateFunc defaultValue curr
 
-    /// Update the TitleView of a NavigationPage, given previous and current values
-    let internal updateNavigationPageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Xamarin.Forms.NavigationPage) =
+    /// Update the attached NavigationPage.TitleView property of a Page, given previous and current values
+    let internal updateNavigationPageTitleView (prevOpt: ViewElement voption) (currOpt: ViewElement voption) (target: Page) =
```
dsyme: I guess change the name to `updatePageTitleView`?

chrisseed: I left it as it was because the TitleView property is on NavigationPage, but you're right, I'll change it.

TimLariviere:
I think it would be better to put the json binding of TitleView in the `attached` array of NavigationPage, since it's an attached property.
Just like
https://github.com/fsprojects/Fabulous/blob/64e6fbe8e19c6ef631f20d3529909af1c3008eb0/tools/Generator/Xamarin.Forms.Core.json#L1515-L1520

Adding it to `Page` is permitted by the Generator, but it can be a problem when someone tries to subclass Page. (they will need to reimplement this property in a wrapper)

TimLariviere:
Retriggering the build, as Travis likes to fail today...

TimLariviere:
Looking good! Glad you were able to extend the Generator. :)

I was hoping this would fail the unit tests, but I guess I didn't covered the `attached` part...
If you feel up to it, you can add new tests for your modifications in https://github.com/fsprojects/Fabulous/blob/master/tests/Generator.Tests/CodeGeneratorTests.fs

>BTW not sure what the policy is on squashing commits - should I have done it before pushing or does it happen when (or should I say if!) the PR is accepted?

We don't have a policy, as far as I know. Maybe that's something to think about.
For now, we accept any commit, be it merged, squashed or rebased. 

Accepting the PR only adds a merge commit to master.

chrisseed:
@TimLariviere Damn - that's because I have Trailing Whitespace Visualizer installed and set to remove whitespace on save and I forgot to only stage and commit the non-whitespace changes.

chrisseed:
Going to start new PR with squashed commits, no whitespace changes and rebased onto current master.

chrisseed:
See #305.

PR #285, opened 2019-01-19
Enable CI build on Azure DevOps

mallibone:
These changes enable building Fabulous on Azure DevOps as discussed in the issue [269](https://github.com/fsprojects/Fabulous/issues/269). Tests are executed on macOS. The builds where executed on a fork: [Build Status?branchName=AzureDevops)](https://dev.azure.com/gnabber/Fabulous/_build/latest?definitionId=22&branchName=AzureDevops)

* Includes the build definition as yaml file
* Adjustments to paths due to differences in the path where .Net Core is installed on ADO

The Linux build is currently still *disabled* due to some issues with the Generator project.

PR #297, opened 2019-01-25
Switching from MSTest to FsUnit in Fabulous.Cli.Tests

TimLariviere:
In order to support tests reporting with #285, MSTest has been removed in favor of FsUnit.
That way, the 2 unit tests projects use the same framework.

It shouldn’t have any impact.


mallibone:
Looks good 👍 after merge I will see to it that we will receive the NUnit results on ADO #285

PR #300, opened 2019-01-26
Fix for assembly name typo in macOS project

sergey-tihon:
Fix for Android joke =)


TimLariviere:
Someone copy/pasted too fast it seems. ;)
Thanks for the fix!

PR #303, opened 2019-01-30
take out fscd, use paket to acquire it

dsyme:
This takes out the core code-format and interpreter code and tests, including it via source inclusion for the moment.


dsyme:
See https://github.com/fsprojects/Fabulous/issues/302

dsyme:
@TimLariviere Just to mention that I still think Fabulous should include its own CLI tool, even if only a very thin packaging wrapper over the generic FSharp.Compiler.PortaCode functionality.

This is because the on-device interpreter in Fabulous.LiveUpdate needs to match the PortaCode data format being sent by the live update tool.  It's much easier for these to be kept in sync if the Fabulous CLI version numbers match the Fabulous.LiveUpdate version numbers (and the Fabulous docs explain that this needs to be the case)

TimLariviere:
@dsyme Ok, I see.
Do we check at runtime if the version matches between the CLI and LiveUpdate?

I think we also need to update the docs to reflect the new to update the CLI when migrating to a new release.

dsyme:
> Do we check at runtime if the version matches between the CLI and LiveUpdate?

No we don't, we should.  I suppose that has to be in the info sent to the webhook

dsyme:
@TimLariviere This is ready

dsyme:
@TimLariviere I've already made some interpreter improvements in FSharp.PortaCode which we should bring in, but we can do that bit by bit

dsyme:
Improvements to the interpreter are here: https://github.com/fsprojects/FSharp.Compiler.PortaCode/pull/2

paket.dependencies
```diff
@@ -3,6 +3,13 @@ group neutral
   framework: netstandard2.0
   source https://api.nuget.org/v3/index.json
   source https://www.myget.org/F/oxyplot
+  github fsprojects/FSharp.Compiler.PortaCode:master src/ProjectCracker.fs
```
TimLariviere: Now that those files evolve independently of Fabulous, maybe we need to force a specific commit rather than `master`?

dsyme: The lock file has the specific commit - if a future `paket update` causes problems that are difficult to solve then at that point we should pin in the dependencies (and then do further updates step by step). Until then tracking master on each `paket update` is ok - the Fabulous tests still test everything all over again.

Issue #283, opened 2019-01-18
Unit testing `Cmd.ofAsyncMsg`


ghost:
My team is working on an app using `Fabulous` framework, and we want to create *unit tests* for each module (not just the root view, update, but children as well) that will test the business logic of the application. But there seem to be certain limitations around testing `Cmd`s, especially no simple work around seems to be available for commands created with `Cmd.ofAsyncMsg`.

Update:
It feels like the right way to test is to create a separate program for each module that is tested and run the program, but it seems like this is **to big** of a setup and using a simple `Cmd` primitive could still be useful.

The main issue with this approach is that in order to get the update loop going we need to create a `ProgrammRunner`, which takes in an instance of a `Xamarin.Forms.Application`, which in our setup we don't want to have `Xamarin.Forms` as a dependency at all. Keep in mind the tests will be run in the host environment, Windows `x86_64`.

Our setup (`FsUnit + xUnit`)
* Shared (Fabulous here)
* Android / iOS
* Shared.Test

What we were able to implement:
```F#
Module.update (Module.Msg.EntryChanged "lol") initModel
  |> fst
  |> should equal "lol"
```

What we need:
```F#
let customDispatch = function
      Ok result -> should be true
      Error error -> failwith "async command failed")
Module.update Module.Msg.SavePressed initModel
  |> snd
  |> Cmd.dispatch customDispatch
```

My questions: 
1. Are there any suggestions about how this custom custom dispatch should be implemented? Am I even taking the right approach to unit testing a `Fabulous` app?
3. Functional programming is unit test - friendly, also any modern application will issue a lot of `async` operations, so why does not `Fabulous` does not talk about it in its documentation as well as does not provide some primitives for doing this type of testing (I might just be a noob that did not read the docs carefully lol :D )

Primitive API proposal for (3):
```F#
// create Cmd's queue
Cmd.ofAsynMsg (HttpGet "http://fsharp.com")
  // dispatch Cmd using custom primitive (this one will wait for all commands to produce a message)
  |> Cmd.dispatch ProgramDispatch<Msg>.TestSynchronousDispatch
  // this test will then verify that the commands indeed resolved to the right messages
  |> should equal [ (Ok "responseBody") ]
```

ps. Thanks for the framework!

TimLariviere:
@kirillpadx I'm glad to hear that you and your team are using Fabulous. :)

We're lacking on the documentation, especially regarding unit testing.
So it's definitely something we'll work on in the future.

[...]

PR #309, opened 2019-02-03
Testable Cmds - Proposal 1 - Added another step to allow bypassing dispatch later

TimLariviere:
This PR is a proposal to solve #283 

Testing Cmd is quite hard today, because it encapsulates its logic inside a function created on the fly which can't easily be unit tested.
Also this function might be asynchronous, returning `unit` early, preventing to know when it actually finishes.

So to solve this problem, I've added a new module `Fabulous.Core.Cmd.Testing`.
This module contains 2 helpers: `executeUntil` and `execute`.
These functions execute synchronously the Cmds to aggregate their generated Msgs, allowing to unit test them.

But synchronously waiting for the Cmd to dispatch a message might end up in a deadlock due to some Cmd not dispatching sometimes (like Cmd.toMsgOption and Cmd.toAsyncMsgOption).
To prevent that, I've changed the signature of `dispatch` to always dispatch an optional message (not changed for `view` though).

For example:

```fsharp
let timerCmd() =
     async {
         do! Async.Sleep 1000
         return TimedTick
     }

let update msg model =
     match msg with
     | TimedTick ->
         let newCount = if model.TimerOn then model.Count + 1 else model.Count
         let newCmd = if model.TimerOn then Cmd.ofAsyncMsg timerCmd() else Cmd.none
         { model with Count = newCount }, newCmd

[<Test>]
let testTimedTick() =
    let initialModel = { Count = 0; TimerOn = true } 
    let actualModel, actualCmd = App.update TimedTick initialModel

   let actualCmdMsg = Cmd.Testing.execute actualCmd

   actualModel |> should equal { Count = 1; TimerOn = true }
   actualCmdMsg |> should equal [ TimedTick ]
```

Here we want to check that if `TimerOn` is true, then we return `timerCmd()`. If false, then `Cmd.none`.

@dsyme Do you think it's a viable solution?
Is it a good use of ValueOption?

I was also thinking of adding some metadata to Cmd in order to provide helpers that could inspect those, instead of actually executing the Cmd.
Maybe it would be a better solution?

### DIFF
--- a/src/Fabulous.Core/ElmishCmd.fs
+++ b/src/Fabulous.Core/ElmishCmd.fs
@@ -2,7 +2,7 @@
 namespace Fabulous.Core
 
 /// Dispatch - feed new message into the processing loop
-type Dispatch<'msg> = 'msg -> unit
+type Dispatch<'msg> = 'msg voption -> unit
 
 /// Subscription - return immediately, but may schedule dispatch of a message at any time
 type Sub<'msg> = Dispatch<'msg> -> unit
@@ -19,18 +19,25 @@ module Cmd =
 
     /// Command to issue a specific message
     let ofMsg (msg:'msg) : Cmd<'msg> =
-        [fun dispatch -> dispatch msg]
+        [fun dispatch -> dispatch (ValueSome msg) ]
 
     /// Command to issue a specific message, only when Option.IsSome = true
     let ofMsgOption (msg:'msg option) : Cmd<'msg> =
         [ fun dispatch ->
              match msg with
-             | None -> ()
-             | Some msg -> dispatch msg ]
+             | None -> dispatch ValueNone
+             | Some msg -> dispatch (ValueSome msg) ]
 
     /// When emitting the message, map to another type
     let map (f: 'a -> 'msg) (cmd: Cmd<'a>) : Cmd<'msg> =
-        cmd |> List.map (fun g -> (fun dispatch -> f >> dispatch) >> g)
+        cmd |> List.map (fun g ->
+            let fOpt value =
+                match value with
+                | ValueNone -> ValueNone
+                | ValueSome x -> ValueSome (f x)
+
+            (fun dispatch -> fOpt >> dispatch) >> g
+        )
 
     /// Aggregate multiple commands
     let batch (cmds: #seq<Cmd<'msg>>) : Cmd<'msg> =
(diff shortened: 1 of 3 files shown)


dsyme:
Hi! I did take a look but didn't fully understand it.  Perhaps a few more comments in the code - does ValueNone to the Cmd dispatch represent an error condition in the unit test?  Or...? thanks

TimLariviere:
@dsyme Ok, I will add more comments to the code.
Essentially what I did was just moving the responsibility to choose or not to dispatch to the main loop from the `Cmd.of...` helpers to `ProgramDispatch<'msg>.dispatch`.

Instead of having
```fsharp
/// Command to issue a specific message, only when Option.IsSome = true
let ofMsgOption (msg:'msg option) : Cmd<'msg> =
    [ fun dispatch ->
            match msg with
            | None -> ()
            | Some msg -> dispatch msg ]
```

We have
```fsharp
/// Command to issue a specific message, only when Option.IsSome = true
let ofMsgOption (msg:'msg option) : Cmd<'msg> =
    [ fun dispatch ->
            match msg with
            | None -> dispatch ValueNone
            | Some msg -> dispatch (ValueSome msg) ]
```

And then `ProgramDispatch<'msg>.dispatch` calls the dispatch implementation if there's something to dispatch.
```fsharp
static let dispatch = 
    id (fun msgOpt -> 
        match msgOpt with
        | ValueNone -> ()
        | ValueSome msg -> dispatchImpl msg)
```

That way, Cmd always calls its callback even if there is no message to dispatch.

>does ValueNone to the Cmd dispatch represent an error condition in the unit test? Or...?

No, ValueNone represents the intent to not dispatch a message.

So for example:
```fsharp
let doSomethingMaybe flag =
     if flag = true then
        Some OtherMessage
     else
        None

let update msg model =
     match msg with
     | SomeMessage flag -> model, Cmd.ofMsgOption (doSomethingMaybe flag)
```

For flag = true, `Cmd.Testing.execute` will return a list of 1 message : `[ OtherMessage ]`
For flag = false, `Cmd.Testing.execute` will return an empty list : `[ ]`

Without this PR, we can't know if the Cmd will ever return something in case of flag = false (especially with Cmd.ofAsyncMsgOption)

Hope it's clearer.

TimLariviere:
Closing this PR because Fabulous will most likely not change the way Cmd behaves today (to stay in line with most Elm-ish implementations)

https://github.com/fsprojects/Fabulous/pull/320#issuecomment-491527076
As this comment suggests, this may be solved by following a simple architectural pattern (applicable at the will of the developer)

PR #312, opened 2019-02-04
[WIP] Added support for Menu

### LINKED ISSUE #252: Add Xamarin.Forms.Menu support
``MenuItem`` is already supported, but it would be nice if we could also build ``Menu``s using the View module.

SergejDK:
With this it should be possible to use Menu in Fabulous.

Items are currently not working because `Items cannot be set`.

fix #252


SergejDK:
1. The problem with the `Items cannot be set` can be cleared if the `Add` and/or `Remove` function for the menu is called. This can be easily repaired.
Example: 
```fsharp
match prevItemsOpt, currItemsOpt with
        | ValueSome prevValue, ValueSome currValue when prevValue = currValue -> ()
        | _, ValueSome currValue -> currValue |> Array.iter(fun value -> target.Add(value))
        | ValueSome prevValue, ValueNone -> prevValue |> Array.iter(fun value -> target.Remove(value) |> ignore)
        | ValueNone, ValueNone -> ()
```

2. Another problem is that it should only be allowed to use `Xamarin.Forms.Menu` to add to a menu. Here it should be needed to transform a ViewElement to the menu, if I see this right. 
Only problem - I don't know how....

TimLariviere:
I think you can take example on `updateListViewItems`.
It does all the things we need here: ObservableCollection, update through Clear/Add/Remove, instantiate ViewElement to controls, etc.
https://github.com/fsprojects/Fabulous/blob/001691db903622673a91cf816d1675d866a3a9a8/src/Fabulous.Core/ViewConverters.fs#L351-L360

SergejDK:
@TimLariviere thanks for the hint! :) 
I will take a look when I get time.

SergejDK:
Need to check the `updateFunc`. 
The calls for the `updateCollectionGeneric` should be false so I need to check it.

SergejDK:
Oh I see you finished it. 
Sorry for the inconvenience that I didn't work on this.. just didn't had the time.

TimLariviere:
It's okay. Just wanted to make this PR part of the next release.
But there's a complication, it does not work currently.

@jwosty @SergejDK Do you have an example how `Xamarin.Forms.Menu` is supposed to be used?
I only found examples of `MenuItem` with `Cell.ContextActions` (that we're missing too).
`Menu` is not a view, so we can add it to a StackLayout or any other layouts.

SergejDK:
Ok, good to know :) Thanks for your hard work on here!


As far as I know it is only used in the `Cell.ContextActions` so in the `ListView` but I never used it when I was working with XF in C#.

SergejDK:
I could think of that `Menu` is used for sth. like a menubar. So e.g. like the menubar at top on the mac.

jwosty:
I thought it was supposed to be used for app menus, but yeah, I can't find that anywhere in XF forms (I thought it just wasn't implemented yet in XF for MacOS).

Because of that, I've written renderers for myself to make it an actual app menu bar on Windows and Mac platforms, and consuming those classes in fabulous is quite verbose. Perhaps it would be useful for me to expose those to the general public as a library :)

jwosty:
@SergejDK at least on MacOS, I haven't observed calling SetMenu on anything (even the App object) doing anything. Maybe it's an example of feature-incompleteness on XF and would merit a discussion there.

TimLariviere:
Found how it's supposed to be used
https://docs.microsoft.com/dotnet/api/xamarin.forms.element.setmenu?view=xamarin-forms#Xamarin_Forms_Element_SetMenu_Xamarin_Forms_BindableObject_Xamarin_Forms_Menu_

SergejDK:
Yeah you can set it to everything but it does not seem to appear in any way. I would think that somehow there will be a menu somewhere... Tested it on WPF/UWP/MacOs with XF but nothings shown. Maybe you will find out more.

jwosty:
Here's a possibly relevant thread from which to glean some information:

https://github.com/xamarin/Xamarin.Forms/pull/4718

SergejDK:
@TimLariviere if in your example the menu is not shown in any way I would say we need to open in issue in the XF repo ?

SergejDK:
Let's keep this here: https://github.com/xamarin/Xamarin.Forms/issues/5344

SergejDK:
@TimLariviere 
will test this PR with 3.6 maybe then the bug is not there anymore....

SergejDK:
Just merged the parts but it is still not working. Even tried my Reproduction Example in XF with C#.

TimLariviere:
@SergejDK Where are we on this one?

SergejDK:
@TimLariviere 
Tried this with XF 4.1 but still does not work. 

Maybe I use the Menu wrong which I dont think but who knows... https://github.com/SergejDK/XFMenuReproduction (is not updated to 4.1)
Do you think we should just close this pr and make a note in the issue?

SergejDK:
@TimLariviere @jwosty 
I'll close this PR and will open a Draft one.

realvictorprm:
so does the menu work meanwhile? :/

PR #313, opened 2019-02-05
Added type annotations on the mkProgram functions

pauldorehill:
When starting to using the `Program.mkProgram`, `Program.mkSimple` the intellisense/compiler currently show generic type names like `'a` which makes it hard to work out what your functions need to do:

image

So I've added the type annotations to make it a bit clearer:

image

### DIFF
--- a/src/Fabulous.Core/ElmishProgram.fs
+++ b/src/Fabulous.Core/ElmishProgram.fs
@@ -31,10 +31,10 @@ type Program<'model, 'msg, 'view> =
       subscribe : 'model -> Cmd<'msg>
       view : 'view
       debug : bool
-      onError : (string*exn) -> unit }
+      onError : (string * exn) -> unit }
 
 /// Starts the Elmish dispatch loop for the page with the given Elmish program
-type ProgramRunner<'model, 'msg>(app: Application, program: Program<'model, 'msg, 'model -> ('msg -> unit) -> ViewElement>)  = 
+type ProgramRunner<'model, 'msg>(app: Application, program: Program<'model, 'msg, 'model -> ('msg -> unit) -> ViewElement>) = 
 
     do Debug.WriteLine "run: computing initial model"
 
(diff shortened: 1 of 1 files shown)


pauldorehill:
As a side note while doing this I was also curious on the `Program` type:
```
type Program<'model, 'msg, 'view> = 
    { init : unit -> 'model * Cmd<'msg>
      update : 'msg -> 'model -> 'model * Cmd<'msg>
      subscribe : 'model -> Cmd<'msg>
      view : 'view
      debug : bool
      onError : (string * exn) -> unit }
```
Does the `view` property need to be generic? I'm just getting started with Xamarin.Forms so there may well be an obvious reason I don't know... but I couldn't see a reason why it can't be defined as:
```
type Program<'model, 'msg> = 
    { init : unit -> 'model * Cmd<'msg>
      update : 'msg -> 'model -> 'model * Cmd<'msg>
      subscribe : 'model -> Cmd<'msg>
      view : 'model -> ('msg -> unit) -> ViewElement
      debug : bool
      onError : (string * exn) -> unit }
```
This would then make the required function signatures clearer still? If you agree I can update the commit?

TimLariviere:
>Does the `view` property need to be generic?
Maybe it's related to the StaticView part, but this seems more like an early code that haven't needed updating since then.
Makes sense to enforce a specific type annotation, clearer that way.

pauldorehill:
Ok, I'll update the pull request with the types updated.

dsyme:
Yes the generic 'view  is related to the half-elmish support 

I wonder if we should duplicate out the half-elmish support into an entirely separate DLL.  Or deprecate/remove it - though there are still people who advocate it as a stepping stone.

pauldorehill:
If you were to remove it I guess the question is `whats the target audience for Fabulous?`. Maybe initially move it to a separate project/dll?

TimLariviere:
Today, most of the work goes to the full-elmish approach. We're not actively supporting half-elmish.
I think we still need to keep it around, but maybe we should go as far as putting it in a different Github project?

That way, we separate it from Fabulous and say we won't support it further.
But if people think it has its place, they will be able to maintain it.

>Maybe initially move it to a separate project/dll? If you want to move I can add as part of this pull request?

I would prefer to merge this PR as-is. It's useful enough.
And do another PR later to add type annotation to `'view`.

TimLariviere:
Merging.
We will separate the half-elmish in a another PR later, before changing type annotation for `'view`

PR #342, opened 2019-02-20
Allow setting the backing classes for UI elements

sdaves:
This small change allows setting ViewBuilders properties and enables me to replace the implementation of the underlying Xamarin.Forms classes used by Fabulous.

For example, if I need to override a virtual method on MasterDetailPage, this change enables running:

```
type MasterDetailPageWithoutToolbar() =
    inherit Xamarin.Forms.MasterDetailPage()
    override __.ShouldShowToolbarButton() = false

Fabulous.DynamicViews.ViewBuilders.CreateFuncMasterDetailPage <- fun () ->
    upcast(new MasterDetailPageWithoutToolbar())
```

This will cause the MasterDetailPageWithoutToolbar class to be loaded when using the standard: 

`View.MasterDetailPage()`

### DIFF
--- a/tools/Generator/CodeGenerator.fs
+++ b/tools/Generator/CodeGenerator.fs
@@ -75,7 +75,7 @@ module CodeGenerator =
         w
 
     let generateCreateFunction (data: CreateData) (w: StringWriter) =
-        w.printfn "    static member val CreateFunc%s : (unit -> %s) = (fun () -> ViewBuilders.Create%s())" data.Name data.FullName data.Name
+        w.printfn "    static member val CreateFunc%s : (unit -> %s) = (fun () -> ViewBuilders.Create%s()) with get, set" data.Name data.FullName data.Name
         w.printfn ""
         w.printfn "    static member Create%s () : %s =" data.Name data.FullName
 
--- a/docs/views-extending.md
+++ b/docs/views-extending.md
@@ -215,6 +215,21 @@ need not be added, you can set them on elements using the helper `With`, usable 
 View.Map(hasZoomEnabled = true, hasScrollEnabled = true).With(horizontalOptions = LayoutOptions.FillAndExpand)
 ```
 
+### Example: MasterDetailPage without a toolbar on UWP with custom ViewBuilders
+
+Fabulous uses ViewBuilders to create the underlying Xamarin.Forms classes. Customizing ViewBuilders is not the recommended way for custom controls but it is a great solution for overridden controls like in the following example:
+
+```fsharp
+type MasterDetailPageWithoutToolbar() =
+    inherit Xamarin.Forms.MasterDetailPage()
+    override __.ShouldShowToolbarButton() = false
+
+Fabulous.DynamicViews.ViewBuilders.CreateFuncMasterDetailPage <- fun () ->
+    upcast(new MasterDetailPageWithoutToolbar())
+
+View.MasterDetailPage() // this now uses MasterDetailPageWithoutToolbar
+```
+
 See also:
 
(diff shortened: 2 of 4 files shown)


tools/Generator/CodeGenerator.fs
```diff
@@ -75,7 +75,7 @@ module CodeGenerator =
         w
 
     let generateCreateFunction (data: CreateData) (w: StringWriter) =
-        w.printfn "    static member val CreateFunc%s : (unit -> %s) = (fun () -> ViewBuilders.Create%s())" data.Name data.FullName data.Name
+        w.printfn "    static member val CreateFunc%s : (unit -> %s) = (fun () -> ViewBuilders.Create%s()) with get, set" data.Name data.FullName data.Name
         w.printfn ""
         w.printfn "    static member Create%s () : %s =" data.Name data.FullName
```
TimLariviere: If I understand correctly, you might want to extend this method instead.
It's the one with the upcast.

TimLariviere:
Ok, I think it could be a good addition.
It's a bit of an advance case, but this will complete our extension points (like ViewProto for default ViewElements).
And it's the same way that Fabulous uses its internal controls instead of the default implementation.

Do you think you can add some documentation on it?
I think we should stress the fact that it's not the recommended way for custom controls, only good for overridden controls like your example.

sdaves:
Sure thing will add some docs. Thanks for the feedback!

tools/Generator/CodeGenerator.fs
sdaves: Yeah, I went with the public setter. I first tried adding a static extension method of the same name in the same namespace hoping that would override the default static methods in ViewBuilders(), but that didn't work.

TimLariviere:
Retriggering CI

sdaves:
Looks like that only retriggered AppVeyor. The Travis build link still says its from 8 hours ago.

SergejDK:
just needs some time.

sdaves:
Awesome it worked :)

TimLariviere:
Could you run `.\build.cmd` (or `.\build.sh` on macOS) to generate the new Xamarin.Forms.Core.fs based on your modification?

sdaves:
Built and uploaded! 
Do you feel this could fit nicely in your release plans for 0.33.0?

TimLariviere:
Thanks!
Yes, it will be released in 0.33.0.

sdaves:
Awesome, looking forward to it 👍

dsyme:
MIght be good to add some docs for ViewProto too? I couldn't spot any on a quick glance

sdaves:
Were you thinking we should document each member in type ViewProtos() or just a doc for the type?

I noticed some items are assigned in type ViewProtos(), but quite a few are left empty, such as ProtoContentPage and ProtoMasterDetailPage. 

ViewProtos members seem to only be called from internal Update methods in the generated `Xamarin.Forms.Core.fs` file, so I'm not sure if they are intended to be used from outside of that file.

TimLariviere:
Ah maybe it's not what I thought it was. If that's the case, it needs to be marked as `internal`.

Nevertheless, I think it would be doable to propose things like this
```fsharp
let view model dispatch =
    View.ContentPage(
        content = View.StackLayout(
            children = [
                View.Label(text="Hello world")
            ]
        )
    )

type App() as app =
   inherit Application()

   ViewBuilders.DefaultLabel <- View.Label(fontFamily="Lobster-Regular")

    let runner =
        Program.mkProgram init update view 
        |> Program.runWithDynamicView app
```

Issue #343, opened 2019-02-21
Elements multiplied after pressing a button


kerams:
Using the SqueakyApp template (v0.32) I built an APK in release mode without making any code changes and installed it on a Onyx Boox Note device (Android 6.0.1). When I press any of the 3 buttons for the first time all of the elements multiply and appear below the original set. It looks like this:

![img_20190221_195253](https://user-images.githubusercontent.com/5063478/53194237-25533780-3613-11e9-8304-669d709e0e92.jpg)

TimLariviere:
@kerams I can't reproduce it. (Android 6.0 / tablet emulator)
Is it specific to this device?

kerams:
I suppose it is. Do you have any tips where to insert logging to get to the bottom of this?

kerams:
Now, I don't know the internals of Fabulous, but is it any similar to Fable/React in terms of updating only the parts of the screen which have changed (by doing some diffing behind the scenes)? I have a feeling there could be something funny going on around those parts.

TimLariviere:
>Now, I don't know the internals of Fabulous, but is it any similar to Fable/React in terms of updating only the parts of the screen which have changed (by doing some diffing behind the scenes)? I have a feeling there could be something funny going on around those parts.

Yes, seems like it. But it's strange to only see it on this device. The same codebase is shared on more than 6 platforms...

>I suppose it is. Do you have any tips where to insert logging to get to the bottom of this?

You can use `Program.withConsoleTrace` (https://fsprojects.github.io/Fabulous/logging.html).
Also Fabulous prints by default in debug output saying when a new control is created.

Other than that, if you can test the CounterApp sample of this repository on your device, maybe take a look at this code below. It's where Fabulous decides if it should reuse a control or create a new one.
https://github.com/fsprojects/Fabulous/blob/e8231902a6cd9b08602e5a9f0dbf87e5dd2b7ef1/src/Fabulous.Core/ViewConverters.fs#L229-L236

kerams:
I'll definitely take a look at that, thanks. What I've found earlier with a minimal repro:

```fsharp
type Model = unit

type Msg = Nothing

let init () = (), []

let update Nothing model = model, []

let view model dispatch =
    View.ContentPage(
        content = View.ListView(verticalOptions = LayoutOptions.Center,
        items = [ 
            View.Button(text = "Increment", horizontalOptions = LayoutOptions.Center, command = fun () -> dispatch Nothing)
[...]

PR #347, opened 2019-02-26
Workaround #343

kerams:
Workaround #343


TimLariviere:
Just thinking about it.
Does it work on your device when deploying as a Release APK?

I fear that the Linker will remove this unused variable.

kerams:
Good point, I'll try it with linking enabled. I was suprised the F# optimizer itself didn't remove it.

kerams:
Can confirm that this workaround works with Sdk and user assembly linking (skipping FSharp.Core, otherwise I get a crash instantly).

PR #348, opened 2019-02-28
Remove reflection from ViewConverters

kerams:
I was looking at `updateIncremental` and wondered whether the reflection is necessary at all. If we're always looking for an `UpdateIncremental` member, why not simply constrain the parameter to be a subtype of `ViewElement`? Having done this, a bunch of generic type parameters have become unnecessary as well. As a result, view updates should be a tiny bit faster.

Feel free to close this if I am overlooking something and the reflection is there for a reason.


kerams:
I assume I need to fix Xamarin.Forms.Core.json too. I'll do that if you want to proceed with this change.

dsyme:
Looks fine to me!

Issue #242, opened 2018-12-02
Add more simple samples


TimLariviere:
AllControls became more of a playground for testing new features than a real sample.
So before we can release Fabulous 1.0, we need to make simpler samples.

I'm thinking something like the `XAML Controls Gallery` in the Microsoft Store (https://www.microsoft.com/en-us/p/xaml-controls-gallery/9msvh128x2zt) can be a good thing to do.
It would both act as an interactive documentation, show code samples on how to use a control, and demonstrate how to implement a complete application with Fabulous.

SergejDK:
Just to clarify what exactly is needed here...

Should this be a new documentation style like the offical Xamarin.Forms documentation? or for example something like this: [MKDocs](http://readthedocs.sheets.ch/user-guide/writing-your-docs/)

or should we write a program as the `XAML Controls Gallery`.

SergejDK:
There are some parts I would like to do better in the current documentation.

- add a searchbar 
- create a menu for better navigation
- add all missing parts (e.g. Switch) but I need to check what else is missing

When this is done then we can add some examples as in the XF e.g. like the TODO-App and so on.
Maybe we can take some of these existing ones and rewrite them in F#.

The AllControls Example is as you wrote it in another Issue more like a Sandbox( Proof of Concept) so I don't think it would be great to rewrite it.
Instead of rewriting I think that it would be better to create the "XAML Gallery".

So the steps would be:
    1. searchbar in doc
    2. menu in doc
    3. adding missing parts in doc
    4. rewrite some existing projects -> maybe pick some out
    5. create the "Gallery"

1-3 would be great for the first milestone.

4/5 maybe for the 2.0.0?

What do you think about this?

TimLariviere:
Regarding the documentation, it's a reasonable request.
We sure could use a complete doc with every controls that we support. (still without going to too far in details, MSDN is there for that) and a better way to navigate / search.
Can you open another issue for this point ?

I don't have enough rights to change the website itself.
It's handled there: https://github.com/fsprojects/fsprojects.github.io
And it's shared between multiple projects owned by fsprojects.

SergejDK:
The new issue belongs to the link you posted, right?

TimLariviere:
No, I meant a new issue in Fabulous so we can track where we are with the docs.

I want to keep this issue for samples, like adding the Todo app you mentioned.

SergejDK:
For anyone who wants to do a sample app but does not which one... 

[...]

PR #350, opened 2019-03-03
Update to XF 3.5

### LINKED ISSUE #329: Upgrade to Xamarin.Forms 3.5
Now that Xamarin.Forms 3.5.0 is out, we can start work to integrate this new version.
This should fix #155 and #295 as well

SergejDK:
The following list shows which are missing right now. Will work on this the next days.

The questionmark behind few classes there I do not know if they are needed.
Interfaces, enums, abstract classes and static classes are not included.
I think that the Attributes and EventArgs do not need to be included, too ???

@TimLariviere could you maybe provide me with some hints for the classes below and the question above? I appreciate your help!

- [x] BaseShellItem
- [x] CarouselView
- [x] CollectionView
- [x] ItemsView
       ScrolledToRequest
       ScrollTo
- [x] ListItemsLayout - not needed?
- [x] MenuShellItem - not needed?
- [x] SearchHandler:
 	ItemsSource 
 	ItemTemplate 
 	OnClearPlaceholderClicked
 	OnItemSelected
 	OnQueryChanged
        OnQueryConfirmed
- [x] SelectableItemsView
- [x] Shell
        GoToAsync
- [x] ShellContent
- [x] ShellGroupItem
- [x] ShellItem
- [x] ShellNavigationState
- [x] ShellSection

fix #329 

Differences:
[https://www.fuget.org/packages/Xamarin.Forms/3.5.0.129452/lib/netstandard2.0/diff/3.4.0.1009999/](https://www.fuget.org/packages/Xamarin.Forms/3.5.0.129452/lib/netstandard2.0/diff/3.4.0.1009999/)


TimLariviere:
>I think that the Attributes and EventArgs do not need to be included, too ???

Yes, if we can reuse the existing classes, we do.
EventArgs are only a "bag" of data, so no need to wrap it in Fabulous.
Attributes as well, it's more related to making custom controls which is not something that concern Fabulous.

TimLariviere:
I didn't had time to check XF 3.5 in details.
But I can say that we are only interested in actual controls, there's no need to wrap things like `MenuItemCollection`, `NavigableElement`, etc.

SergejDK:
Great! Thanks for your feedback :)!

SergejDK:
I think i got everything...

TimLariviere:
@SergejDK Do you know why CollectionView/CarouselView/Shell are in XF 3.5?
Everything I read about them is about Xamarin.Forms 4.0 Preview

SergejDK:
They are already in 3.5 but only allowed to use when setting: global::Xamarin.Forms.Forms.SetFlags("Shell_Experimental");

[https://docs.microsoft.com/de-de/xamarin/xamarin-forms/app-fundamentals/shell?tabs=android](https://docs.microsoft.com/de-de/xamarin/xamarin-forms/app-fundamentals/shell?tabs=android)

The question why it is in 3.5 I think because of earlier feedback from the devs. 

> BONUS: Keep in mind that 3.5.0 also contains previews of other new features that the Xamarin team is currently developing,. We look forward to receiving your feedback on Shell, Visual, CollectionView, and CarouselView. As you explore these features, please take a moment to share with us your thoughts here.

[https://blog.xamarin.com/5-things-youll-love-xamarin-forms-3-5/](https://blog.xamarin.com/5-things-youll-love-xamarin-forms-3-5/)

SergejDK:
@TimLariviere the fail on travis - is it the error where it only needs to rerun?

TimLariviere:
Seems like it.
For some reasons, it failed to download Xamarin.Forms.
Will retrigger the CI

SergejDK:
@TimLariviere 
Any other ideas? Most times the error goes away when rerun but right now not...
Maybe I missed a dependency?

TimLariviere:
Apparently they changed their targeting framework with Xamarin.Forms 3.5.
In XF <= 3.4, Xamarin.Android was using the folder `MonoAndroid10`
But now, it's using either `MonoAndroid81` or `MonoAndroid90`

image

So the links in both the samples app and the templates are no longer valid.

For example:
```xml
<Reference Include="Xamarin.Forms.Core">
    <HintPath>..\packages/Xamarin.Forms.XamarinFormsSdk/lib/MonoAndroid10/Xamarin.Forms.Core.dll</HintPath>
</Reference>
```
will become
```xml
<Reference Include="Xamarin.Forms.Core">
    <HintPath>..\packages/Xamarin.Forms.XamarinFormsSdk/lib/MonoAndroid90/Xamarin.Forms.Core.dll</HintPath>
</Reference>
```

SergejDK:
Thanks for looking into it! Could not do it when I was at work... 
Will change it.

tools/Generator/Xamarin.Forms.Core.json
```diff
@@ -2014,6 +2040,430 @@
...
+          "name": "GoToAsync",
+          "uniqueName": "ssGoToAsync",
+          "inputType": "string list * Map<string, string> * Fabulous.DynamicViews.AnimationKind",
+          "modelType": "string list * Map<string, string> * Fabulous.DynamicViews.AnimationKind",
+          "updateCode": "(fun _ curr target -> triggerSSGoToAsync curr target)"
+        }
+      ]
+    },
+    {
+      "name": "Xamarin.Forms.CarouselView",
```
TimLariviere: CarouselView and CollectionView have no members on their own?

TimLariviere:
Could you add an example of how to use the new controls in the AllControls samples?

tools/Generator/Xamarin.Forms.Core.json
SergejDK: I just looked at fuget and there were no members. Maybe they will get some in 3.6? 
I check this.

SergejDK: The GitHub classes of those two seems to be the same so it's correct?

SergejDK:
Would it not be better to have a project for shell and not include it in AllControls? 
I think the AllControls will be hard to read if we add everything in it. 

And with the Shell it mabye easier to work on: #242

tools/Generator/Xamarin.Forms.Core.json
TimLariviere: Ah ok. They don't have much code by themselves, it's mostly inherited from ItemsView/SelectableItemsView. Alright.

TimLariviere:
I think we should still add the new controls to AllControls, it's more a smoke test than a sample anymore. That way, we can check the new controls are working and we don't break them later.

Though indeed, Shell would be easier to use in a new project and provides a good structure for new samples.

SergejDK:
I will add simple parts in the allcontrols.

TimLariviere:
Thanks!

SergejDK:
> Existing iOS and Android applications can adopt Shell and benefit immediately from navigation, performance, and extensibility improvements.

Because Shell seems to be only available on Android and iOS I have to ask for the platform. This is just a note that you won't be suprised about the addition to ask for android and ios.

SergejDK:
And if I am right I made more work than I had to do... I think I can delete the BaseShellItem?

SergejDK:
@TimLariviere could you test on Android and iOS for the ShellView? 
I kinda run in a very interesting issue... on iOS I get a `Parameter can not be null` Exception.
On Android I get no exception but it does not display the content.

Will take a closer look tomorrow/today. This could be a problem of XF but I think I messed something up when I included the `ItemsProperty`.

SergejDK:
I did definetly messed up the `updateCode` because on Android I tested with the TicTacToe and it is rendered completely wrong.

SergejDK:
@TimLariviere 
I think for shell we do truly need a project because we need to use sth like.
```fsharp
FormsAppCompatActivity.TabLayoutResource <- TicTacToe.Droid.Resources.Layout.Tabbar
FormsAppCompatActivity.ToolbarResource <- TicTacToe.Droid.Resources.Layout.Toolbar

Else it should give an error.

If I am correct then there is only CarouselView/CollectionView and maybe searchhandler to make examples in allcontrols?

Last thing: I am still trying to figure out what is wrong with the items updateCode. If you have any hint I would appreciate it!!

TimLariviere:
>And if I am right I made more work than I had to do... I think I can delete the BaseShellItem?

Yes, I think you're right.
Some "abstract" controls don't need to be exposed in Fabulous. We only want to expose those we can use directly.

>If I am correct then there is only CarouselView/CollectionView to make examples in allcontrols?

Yes, there seems to be only 3 new controls: Shell, CarouselView and CollectionView.
The others are more generic controls that serve as a base for those 3.

SergejDK:
Added examples for carousel and collectionview but I think there is something wrong.
If I try to add sth. like `View.Label` in the `itemssource` property it won't render.
So here I think I need to overwrite it somehow in the members or is there another way?

TimLariviere:
@SergejDK The `ItemsSource` property on `ItemsView` (inherited by Carousel and CollectionView) is not enough for Fabulous. ViewElement won't render itself by default.

You need to do something like the `ListView` does: 
- Create a data holder with PropertyChanged support
https://github.com/fsprojects/Fabulous/blob/1c435ccc61d58efb6820706bce56c24e68b00a28/src/Fabulous.Core/ViewConverters.fs#L27-L41
- Create a custom made control which will be responsible of rendering the related ViewElement
https://github.com/fsprojects/Fabulous/blob/1c435ccc61d58efb6820706bce56c24e68b00a28/src/Fabulous.Core/ViewConverters.fs#L70-L112
- Create a custom CarouselView/CollectionView initialized with a DataTemplate of the new control (which is responsible for rendering the ViewElement)
https://github.com/fsprojects/Fabulous/blob/1c435ccc61d58efb6820706bce56c24e68b00a28/src/Fabulous.Core/ViewConverters.fs#L114-L116
- And finally, tell the Generator you want to instantiate your custom control whenever you call `View.Carousel`/`View.CollectionView`
https://github.com/fsprojects/Fabulous/blob/1c435ccc61d58efb6820706bce56c24e68b00a28/tools/Generator/Xamarin.Forms.Core.json#L1804-L1814

That way, when you add a `View.Label` in the ItemsSource property, Fabulous will wrap it in a `ItemsViewElementData` (1st step) and add it to the actual ItemsSource property.
Then when displaying, the Carousel/CollectionView will call `ItemsViewItem` to handle the rendering of the `ItemsViewElementData`.

SergejDK:
@TimLariviere great explanation! Thanks!

As far as I see I can nearly copy everything except changing some names and inherited classes. Am I right?
-- Tried it that way but I get an InvalidCastException. Currently I am searching where this happens

TimLariviere:
>As far as I see I can nearly copy everything except changing some names and inherited classes. Am I right?

Yes, it should mostly work as-is.
Just watch out for ` inherit ViewCell() `, Carousel/CollectionView might not use this class for templating.

SergejDK:
It seems like everythign can be used. 
[https://blog.xamarin.com/xamarin-forms-4-0-feature-preview-an-entirely-new-point-of-collectionview/](https://blog.xamarin.com/xamarin-forms-4-0-feature-preview-an-entirely-new-point-of-collectionview/)

So maybe inherit from view directly?  The difference to ListView is that listview uses `ItemsView<Cell>` but Collection/Carousel uses `ItemsView` class.

TimLariviere:
>So maybe inherit from view directly?

@SergejDK You can inherit from ContentView and append the ViewElement's control to the `Content` property.

SergejDK:
Seems to work but it does not display anything. It gets created correctly though...

SergejDK:
The Labels seems to be created but not shown in the ui:
grafik

Is there still sth. wrong within `ItemViewElementCell`?
No Exception - nothing to check... Any hint on this?

SergejDK:
Got time again.
Very interesting... 
CollectionView on iOS is working on Android not.
CarouselView does not work. Will take a deeper look

SergejDK:
@TimLariviere 
if I get it right the `CarouselView` for iOS is not implemented in XF3.5: [https://github.com/xamarin/Xamarin.Forms/issues/5488](https://github.com/xamarin/Xamarin.Forms/issues/5488)
So this won't work no matter how much effort I put into it.

will test for android part.

SergejDK:
@TimLariviere 
it can be that with further updates maybe to 3.6/4.0 the issue won't be there anymore.

So maybe it is okay to merge here even though the examples for the 3 parts are currently not working. 

For those example we can open each issue and work on the updates so we can catch up there and afterwards test the examples again. 

What do you think?

TimLariviere:
@SergejDK Alright.
As those controls are preview only in XF 3.5, I think we can merge this PR for now and fix them once they're fully available (likely with XF 4.0).

I'll add a note for these controls in the docs saying that while we provide an initial support, there may be some issues with them.

SergejDK:
@TimLariviere 
Great Job! Thanks for keeping up with the work!

SergejDK:
@TimLariviere 
I think it is the normal problem where we need to restart the build on travis?

TimLariviere:
@SergejDK Thank you for your work in this long running PR. :)
I'm merging this now.

This evening I'll be releasing this in a new version on NuGet.

SergejDK:
@TimLariviere 
always happy to help :)
I think everyone is waiting for the release

what do you think about making the step to 3.6 ? Maybe I can make it before evening and you could release it with 3.6?

aspnetde:
> what do you think about making the step to 3.6 ? Maybe I can make it before evening and you could release it with 3.6?

That would be awesome!

PR #360, opened 2019-03-18
Updated the validation example

pauldorehill:
While there are many ways to skin a cat when it comes to validation.... I thought the example given here for validation could be improved. Currently it doesn't carry forward the original input value along with the error message (you would likely need both of these); and it also in a way reinvents the result type on the `Temperature` union type rather than wrapping the temperature in a `Result` (which would have advantages when you had a large form of values and you wanted to check if they were all valid). Let me know your thoughts.

### DIFF
--- a/docs/models.md
+++ b/docs/models.md
@@ -8,36 +8,30 @@ Models
 
 ### Messages and Validation
 
-Validation is generally done on updates to the model, storing error messages from validation logic in the model
-so they can be correctly and simply displayed to the user.  Here is an example of a typical pattern.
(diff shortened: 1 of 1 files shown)


TimLariviere:
Yes, the original example was a little simplistic, but it's good for beginners or people who want to quickly check how to do validation in Fabulous.

Your example is good too, but harder to read (it does more things).

So maybe we should add it as a second example under an "advanced/real life" section?

pauldorehill:
Ok sounds good. I'll put something together... might be better served as an example app?

TimLariviere:
@pauldorehill I think it would be best to add it to the docs in the first time, like you did.
People tend to check the docs first and not necessarily the samples.

pauldorehill:
Tim, I've updated the commit based on your comments

PR #371, opened 2019-03-31
Tell users to add the Oxyplot MyGet feed as a NuGet restore feed

Happypig375:
The Xamarin Oxyplot packages are not likely to be further updated given https://github.com/oxyplot/oxyplot/issues/1274, so why not give clearer instructions based on the current situation?

### DIFF
--- a/docs/views-oxyplot.md
+++ b/docs/views-oxyplot.md
@@ -9,7 +9,14 @@ Using OxyPlot Charts
 Below is an example of an extension for [OxyPlot](http://docs.oxyplot.org/). To use the extension:
 
 1. Follow the instructions to [add references and initialize renderers](http://docs.oxyplot.org/en/latest/getting-started/hello-xamarin-forms.html)
-2. Add a reference to the [`Fabulous.OxyPlot`](https://www.nuget.org/packages/Fabulous.OxyPlot) package across your solution.
+2. `Fabulous.Oxyplot` references a version of `Oxyplot.Xamarin.Forms` that is on the Oxyplot MyGet feed instead of NuGet.
+Therefore, add these lines inside your cross-platform .NET Standard `.fsproj` file:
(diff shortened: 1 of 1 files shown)


TimLariviere:
@Happypig375 Would it make sense to downgrade the package we use, and only target nuget.org?

Happypig375:
[Diff of oxyplot/oxyplot from 1.1.0-unstable0011 to 1.1.0-unstable0013](https://github.com/oxyplot/oxyplot/compare/5bbea0be62d2cc1ca40dd06212400af25b177c3a..ddd083897dc124be17d120a3db63ae565c68aaa7)
[Diff of oxyplot/oxyplot-xamarin from 1.1.0-unstable0011 to 1.1.0-unstable0013](https://github.com/oxyplot/oxyplot-xamarin/compare/e0ca6f98442074fdeb0b5240f7209cbb05ea26a7..b4cfbdefd43e3d53ddbe14a7d231e0c2ded03e6c)

I don't think there's much of a difference in Oxyplot.Xamarin.Forms 1.1.0-unstable0013 apart from requiring an unstable version of Oxyplot.

Happypig375:
There might be something overlooked by me though.

TimLariviere:
@Happypig375 Ok, great. Will try to see if the samples run with `1.1.0-unstable0011`.
If there's no issue, I suggest we downgrade.
It will be easier for everyone.

What do you think?

Happypig375:
We might be able to downgrade even further to `1.0.0`, the only stable version on NuGet. That way, we will be able to potentially remove undiscovered bugs due to an "unstable" version.

TimLariviere:
Created PR #387 with the downgrade.
Building the source code is OK, though I haven't tested it in an app.

Happypig375:
Superseded by #387

PR #406, opened 2019-05-16
XF Update 3.6

### LINKED ISSUE #357: Upgrade to Xamarin.Forms 3.6
Xamarin.Forms 3.6 has been released. Fuget API changes since 3.5 [here](https://www.fuget.org/packages/Xamarin.Forms/3.6.0.220655/lib/netstandard2.0/diff/3.5.0.169047/).

SergejDK:
fix #357 

2 Functions are open right now the rest is in it. 
On the Slider class: `DragStarted` and `DragCompleted`

### DIFF
--- a/Packages.targets
+++ b/Packages.targets
@@ -2,9 +2,9 @@
     <ItemGroup>
         <PackageReference Update="FSharp.Core" Version="4.5.2" />
         <PackageReference Update="FSharp.Data" Version="3.0.0" />
-        <PackageReference Update="Xamarin.Forms" Version="3.5.0.129452" />
-        <PackageReference Update="Xamarin.Forms.Platform.WPF" Version="3.5.0.129452" />
-        <PackageReference Update="Xamarin.Forms.Maps" Version="3.5.0.129452" />
+        <PackageReference Update="Xamarin.Forms" Version="3.6.0.344457" />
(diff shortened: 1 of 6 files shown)


aspnetde:
One question: Did all the visual elements already get the `visual="..."` attribute with the 3.5 update? I can't tell from the changeset. This would be necessary to apply the [Material Visual](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/visual/material-visual).

SergejDK:
I don't think that in 3.5 there was the attribute. 
This should be in 3.6 - will test it in the `AllControls.fs`.

I don't know if we need an Flag here as for Shell etc.

SergejDK:
@aspnetde 
Normally this was included in 3.5 but I didn't see that. I will include it with this update.

aspnetde:
Thanks a lot. Just in time for me (really).

SergejDK:
@TimLariviere I think I got it all.

TimLariviere:
@SergejDK Well, that was quick. 😄
I checked and everything is covered. Great job!

I will integrate it to v0.34 then :)

SergejDK:
@TimLariviere 
sure the guys are waiting so we need to be quick :D

Issue #161, opened 2018-08-15
Infinite loop when messages are sent too quickly in succession


TimLariviere:
**Applies to:** Android (maybe iOS)

**Description:**
To update the model along with what the user does, I listen to events like `TextChanged`.
This works fine.
But under some circumstances, such event might trigger an infinite update-view loop which will freeze the app.
Based on what I do and the logs, I suspect EXF doesn't like TextChanged to trigger too quickly in a row.
I will update this issue when I find more about this.

iOS seems not to be affected.

**Logs**
In ElmishContacts, I searched for a contact "Hoya Booya" and then erased my search.
```
Message: MainPageMsg (TabAllContactsMsg (UpdateFilterText "Ho"))
Updated model: ...
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Label
View, model = ...
View result: NavigationPage(...)@821839065
Updating NavigationPage, prevCount = 1, newCount = 1
Adjust page number 0
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Image
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Label
Create Xamarin.Forms.Label
Create Xamarin.Forms.Image

Message: MainPageMsg (TabAllContactsMsg (UpdateFilterText "H"))
Updated model: ...
View, model = ...
View result: NavigationPage(...)@571913183
Updating NavigationPage, prevCount = 1, newCount = 1
Adjust page number 0
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Label

Message: MainPageMsg (TabAllContactsMsg (UpdateFilterText ""))
Updated model: ...
View, model = ...
View result: NavigationPage(...)@936190227
Updating NavigationPage, prevCount = 1, newCount = 1
Adjust page number 0
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Label
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Image
Create Xamarin.Forms.StackLayout
Create Xamarin.Forms.Label
Create Xamarin.Forms.Label

[IInputConnectionWrapper] beginBatchEdit on inactive InputConnection
[IInputConnectionWrapper] getTextBeforeCursor on inactive InputConnection
Create Xamarin.Forms.Image
Message: MainPageMsg (TabAllContactsMsg (UpdateFilterText "H"))
Updated model: ...
View, model = ...
View result: NavigationPage(...)@906179063
[...]

PR #414, opened 2019-05-19
[WIP] Prevent programmatic changes to trigger events

### LINKED ISSUE #304: TextChanged is incorrectly fired when changing the model
The event is fired even though the user has not input anything when the model is switching.

```fsharp
type Model = 
    {
        ShowGreeting : bool
        Value : string
    }

type Msg = 
    | ShowGreeting
    | ShowValue
    | SetValue of string

let initModel = { ShowGreeting = false; Value = "text" }

let init () = initModel, Cmd.none

let update msg model =
    match msg with
    | ShowGreeting -> { model with ShowGreeting = true }, Cmd.none
    | ShowValue -> { model with ShowGreeting = false }, Cmd.none
    | SetValue v -> { model with Value = v}, Cmd.none

let view (model: Model) dispatch =
    let text = if model.ShowGreeting then "Hello, World!"
               else model.Value
    let onTextChanged (e : TextChangedEventArgs) =
        printfn "TEXT CHANGED from:%s to:%s" e.OldTextValue e.NewTextValue
        e.NewTextValue |> SetValue |> dispatch
    View.ContentPage(
      content = View.StackLayout(padding = 44.0,
        children = [ 
            View.Button(text = "Show Greeting", command = (fun () -> dispatch ShowGreeting))
            View.Button(text = "Show Value", command = (fun () -> dispatch ShowValue))
            View.Entry(text = text, textChanged = onTextChanged)
        ]))
```

Repro:

1. Run the given app
2. Note it says "text" because ShowGreeting = false
3. Tap "Show Greeting" and note correct text
4. Tap "Show Value" and not **incorrect text** it show "Hello, World!" instead of "text"

This is because the `ShowGreeting` forced a `SetValue` message. This can be seen in the trace:

```
Message: ShowGreeting
Updated model: {ShowGreeting = true; Value = "text";}
View, model = {ShowGreeting = true; Value = "text";}
View result: ContentPage(...)@-1009560494

TEXT CHANGED from:text to:Hello, World!  <-- Should not be called

Message: SetValue "Hello, World!"
Updated model: {ShowGreeting = true; Value = "Hello, World!";}
View, model = {ShowGreeting = true; Value = "Hello, World!";}
View result: ContentPage(...)@-1762026536
```

The bug is in `UpdateEntry` (and `UpdateSearchBar`, `Editor`). The `Text` property is set before the `TextChanged` event. This causes `TextChanged` to fire using old model data.

https://github.com/fsprojects/Fabulous/blob/a16e11eeadf6c70479b38315f31ffa7c65e0ff83/src/Fabulous.Core/Xamarin.Forms.Core.fs#L7508-L7564

To fix it, `TextChanged` should be unsubscribed from before setting `Text`. In general, this pattern should be followed for all `PropertyChanged` events (I have not audited the code). The event should first be unsubscribed, then the property set, then re-subscribed if needed.

I ran into this bug while building a data entry UI in my app.

TimLariviere:
>// WIP : Need better checks to avoid unsubscribing/resubscribing for nothing; Also need unit tests and docs

Closes #304

All events related to property change in Xamarin.Forms are triggered both by UI and programmatic updates.
While events triggered by programmatic changes are acceptable in OOP, in Fabulous, there is some drawbacks:
- Unnecessary update messages (model is already up-to-date, no need to update it again)
- Unnecessary UI updates
- In some conditions, this may trigger infinite loops (see #161)
- In some cases, this might introduce unwanted behaviors in Fabulous (see #304)

Overall this systematic event triggering has a negative impact on both performance and ease of use (requires developers to use workarounds like `debounce` to reduce the number of messages)

This PR changes the way DynamicViews is generated, going away from strict XF matching. Events are now handled separately.
Instead something like that (which would trigger a `TextChanged` event when updating `Text`)
```fsharp
match prevTextChangedOpt, currTextChangedOpt with
| ValueSome prevValue, ValueSome currValue when identical prevValue currValue -> ()
| ValueSome prevValue, ValueSome currValue -> target.TextChanged.RemoveHandler(prevValue); target.TextChanged.AddHandler(currValue)
| ValueSome prevValue, ValueNone -> target.TextChanged.RemoveHandler(prevValue)
| ValueNone, ValueSome currValue -> target.TextChanged.AddHandler(currValue)
| ValueNone, ValueNone -> ()

match prevTextOpt, currTextOpt with
| ValueSome prevValue, ValueSome currValue when identical prevValue currValue -> ()
| _, ValueSome currValue -> target.Text <- currValue
| ValueSome _, ValueNone -> target.Text <- ""
| ValueNone, ValueNone -> ()
```

This was rewrote like:
```fsharp
let shouldUpdateTextChanged = not ((identical prevTextChangedOpt currTextChangedOpt) && (identical prevTextOpt currTextOpt))

if shouldUpdateTextChanged then
    match prevTextChangedOpt with
    | ValueSome prevValue -> target.TextChanged.RemoveHandler(prevValue)
    | ValueNone -> ()

match prevTextOpt, currTextOpt with
| ValueSome prevValue, ValueSome currValue when identical prevValue currValue -> ()
| _, ValueSome currValue -> target.Text <- currValue
| ValueSome _, ValueNone -> target.Text <- ""
| ValueNone, ValueNone -> ()

if shouldUpdateTextChanged then
    match currTextChangedOpt with
    | ValueSome currValue -> target.TextChanged.AddHandler(currValue)
    | ValueNone -> ()
```

### DIFF
--- a/tools/Generator/CodeGenerator.Models.fs
+++ b/tools/Generator/CodeGenerator.Models.fs
@@ -32,7 +32,12 @@ module CodeGeneratorModels =
           TypeToInstantiate : string
           Parameters : string [] }
 
-    type UpdateMember =
+    type UpdateEvent =
+        { Name : string
+          UniqueName : string
+          ModelType : string }        
+
+    type UpdateProperty =
         { Name : string
           UniqueName : string
           ModelType : string
@@ -42,13 +47,14 @@ module CodeGeneratorModels =
           ElementTypeFullName : string option
           IsParameter : bool
           BoundType : BoundType option
-          Attached : UpdateMember [] }
+          Attached : UpdateProperty [] }
 
     type UpdateData =
         { Name : string
           FullName : string
           BaseName : string option
-          ImmediateMembers : UpdateMember []
+          ImmediateEvents : UpdateEvent []
+          ImmediateProperties : UpdateProperty []
           KnownTypes : string [] }
 
     type ConstructData =
--- a/tools/Generator/CodeGenerator.Models.fs
+++ b/tools/Generator/CodeGenerator.Models.fs
@@ -35,7 +35,8 @@ module CodeGeneratorModels =
     type UpdateEvent =
         { Name : string
           UniqueName : string
-          ModelType : string }        
+          ModelType : string
+          RelatedProperties : string [] }        
 
     type UpdateProperty =
         { Name : string
--- a/tools/Generator/CodeGenerator.Preparation.fs
+++ b/tools/Generator/CodeGenerator.Preparation.fs
@@ -16,7 +16,8 @@ module CodeGeneratorPreparation =
           InputType : string
           ModelType : string
           ConvToModel : string
-          IsImmediateMember : bool }
+          IsImmediateMember : bool
+          RelatedProperties : string [] }
 
     type PreparedProperty =
         { Name : string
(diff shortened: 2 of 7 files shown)


TimLariviere:
Closed in favor of #530

PR #418, opened 2019-05-22
Add examples for the new CmdMsg pattern allowing unit testability of Cmds

### LINKED ISSUE #402: Proposal: A more robust way to unit test views
I took at look at the test [examples](https://github.com/fsprojects/Fabulous/blob/master/samples/CounterApp/CounterApp.Tests/Tests.fs#L80) for views and gave it a try. While the `*Viewer` types are very helpful and the overall mechanism works I think it's kind of an anti-pattern on how it is demonstrated here (and in the docs).

Obviously you _can_ build your tests that way, but I don't think you should. Especially not when you're dealing with more complex view hierarchies, and when you're beginning to build your UI which may at that time be subject of change quite often. 

Because **every time the hierarchy changes your tests would break** and require you to update them. Which is a tedious process as you need to scan your hierarchy and replicate it in your tests. So you not only end up with a lot of boilerplate code but very fragile boilerplate code.

### Example

```fs
let loginPage() = 
    View.ContentPage(
        content = View.ScrollView(
            content = View.StackLayout(
                padding = 30.0,
                children = [
                    View.Frame(
                        verticalOptions = LayoutOptions.CenterAndExpand,
                        content = View.StackLayout(children = [
                            View.Entry(
                                placeholder = "User name",
                                isEnabled = (not model.IsBusy),
                                textChanged = (fun args -> (dispatch (LoginUserNameChanged args.NewTextValue))))
                            View.Entry(
                                placeholder = "Password", 
                                isPassword = true,
                                isEnabled = (not model.IsBusy),
                                textChanged = (fun args -> (dispatch (LoginPasswordChanged args.NewTextValue))))
                            View.Button(
                                automationId = "SignInButton",
                                text = "Sign in", 
                                heightRequest = 30.0,
                                isVisible = (not model.IsBusy),
                                command = (fun () -> dispatch LoginStarted),
                                canExecute = model.Login.IsCredentialsProvided)
                            View.ActivityIndicator(
                                isRunning = true,
                                heightRequest = 30.0,
                                isVisible = model.IsBusy)])
                    )
                ]
            )
        )
    )
```

I have that login form and want to make sure the sign in button is enabled only when the `IsCredentialsProvided` property is set to `true`. This is crucial to the whole thing and worth testing (from my point of view).

A test would now look like:

```fs
[<Theory>]
[<InlineData(true)>]
[<InlineData(false)>]
let ``The Sign In Button is only enabled when the Credentials are provided`` credentialsProvided =
    let model = initialModel()
    let model = { model with Login = { model.Login with IsCredentialsProvided = credentialsProvided } }
    
    let view = (App.view model ignore)
    let contentPage = ContentPageViewer(view)
    let scrollView = ScrollViewViewer(contentPage.Content)
    let stackLayout = StackLayoutViewer(scrollView.Content)
    let frame = FrameViewer(stackLayout.Children.[0])
    let stackLayout = StackLayoutViewer(frame.Content)
    let button = ButtonViewer(stackLayout.Children.[2])
    
    button.CanExecute |> should equal credentialsProvided
```

This is not a new problem, though. UI test frameworks seem to agree on unique identifiers within the view hierarchy which they can use to "pick" an element effortlessly. [Xamarin.Forms has those as well](https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.element.automationid?view=xamarin-forms) and is turns out they are even part of the Fabulous "bindings".

I wrote a small helper function that covers the most common scenarios right now (Content, Children, and Pages) and is able to find a given element by its `AutomationId` in the hierarchy:

```fs
/// Looks for an view element with the given Automation ID in the view hierarchy.
/// This function is not optimized for efficiency and may execute slowly.
/// Which shouldn't be noticable in pure test scenarios, though.
let rec tryFindViewElement automationId (element:ViewElement) =
    let elementAutomationId = element.TryGetAttributeKeyed(ViewAttributes.AutomationIdAttribKey)
    if elementAutomationId.IsSome && elementAutomationId.Value = automationId then
        Some(element)
    else
        let childElements = ResizeArray<ViewElement>()
                        
        let content = element.TryGetAttributeKeyed(ViewAttributes.ContentAttribKey)
        if content.IsSome then
            childElements.Add content.Value
        else                  
            let pages = element.TryGetAttributeKeyed(ViewAttributes.PagesAttribKey)
            if (pages.IsSome) then
                childElements.AddRange pages.Value
            else                    
                let children = element.TryGetAttributeKeyed(ViewAttributes.ChildrenAttribKey)
                if (children.IsSome) then
                    childElements.AddRange children.Value
           
        let childElementResult =
            childElements
            |> Seq.map (fun e -> e |> tryFindViewElement automationId)
            |> Seq.filter (fun e -> e.IsSome)
            |> Seq.tryHead
            
        match childElementResult with
        | Some result -> result
        | None -> None
```

My test is now much simpler and doesn't rely on the concrete structure of the form as long as the button still is to be found somewhere:

```fs
[<Theory>]
[<InlineData(true)>]
[<InlineData(false)>]
let ``The Sign In Button is only enabled when the Credentials are provided`` credentialsProvided =
    let model = initialModel()
    let model = { model with Login = { model.Login with IsCredentialsProvided = credentialsProvided } }
    
    let view = (App.view model ignore)
    let button = view |> tryFindViewElement "SignInButton"
    button.IsSome |> should be True
    
    let button = ButtonViewer(button.Value)
    button.CanExecute |> should equal credentialsProvided
```

I'm pretty sure `tryFindViewElement` can be optimized in certain ways (bear with me, I'm a C# developer by day :-)). But maybe someone may find it helpful anyway.

TimLariviere:
Closes #402 (this PR includes the `tryFindViewElement` helper function)

After much discussion, we came to the conclusion that a simple pattern can allow complete unit testing when using `Cmd<'msg>` despite it not being unit testable.
(See full discussion in #320 and #309)

This pattern is completely up to the developer, nothing has been changed in Fabulous to support it.

The fact is that `Cmd<'msg>` is not unit testable because it's essentially just an array of functions.
Functions returning Cmds won't be fully eligible for unit testing, that include `init` and `update`, which would otherwise be perfect for unit testing...

The idea of this pattern is to remove the dependency to `Cmd<'msg>` from `init` and `update` by using an intermediate `CmdMsg` discriminated union.
This union will only state intents instead of actually executing a Cmd.

_Updated with Program.mkProgramWithCmdMsg_

```fsharp
type Msg =
    | FieldUpdated of string
    | FieldValidated of bool

type Model =
    { Text: string
      IsValidated: bool }

// The new discriminated union replacing direct usage of Cmd<'msg>
type CmdMsg =
    | ValidateField of string

// Some long running validation process that will eventually dispatch a Msg.FieldValidated
let validateFieldAsync value = async { ... }

/// unit -> Model * CmdMsg list
let init () =
    { Text = ""; IsValidated = false }, [] // Cmd.none is replaced by an empty list of CmdMsg

/// Msg -> Model -> Model * CmdMsg list
let update msg model =
    match msg with
    | FieldUpdated value -> { model with Text = value }, [ (ValidateField value) ] // Ask to validate the field value - equivalent of Cmd.ofAsync (validateFieldAsync  value)
    | FieldValidated validation -> { model with IsValidated = validation }, []

let view model dispatch = (...)

// A new function that will map a CmdMsg to the actual Cmd to execute
let mapCommands cmdMsg =
    match cmdMsg with
    | ValidateField value -> Cmd.ofAsync (validateFieldAsync value)

// Finally starts the program
type App() =
    Program.mkProgramWithCmdMsg init update view mapCommands
    |> Program.run...
```

`init` and `update` now have a full data-only return type.
Greatly simplifying unit testing.

```fsharp
[<Test>]
let updatingFieldShouldUpdateModelAndTriggerValidation () =
    let initialModel = { Text = "Previous Text"; IsValidated = false }
    let expectedReturn = { Text = "New Text"; IsValidated = false }, [ (ValidateField "New Text") ]
    App.update (FieldUpdated "New Text") initialModel |> should equal expectedReturn
```


TimLariviere:
Interestingly enough, this also completely enable code reuse between Fabulous and Elmish.
`init` and `update` 100% not specific to Fabulous.

TimLariviere:
@cmeeren & @aspnetde : Don't hesitate to give your opinion on this PR

cmeeren:
Looks good! You might want to consider including helper functions that accept a `mapCommands` function and an `update` function and returns a `convertedUpdate` function, and similar for `init`. That way, users won't have to implement the (allegedly trivial) mapping themselves, and it's possible to simply inline it into the `Program.mk...` call.

aspnetde:
I'm fine with it, too.

(Side note: In my small sample app I ended up with a similar approach like the one used in the ElmishContacts sample, having pages with `Msg` + `ExternalMsg` to communicate with their parents. I also had one top level command to show alerts. Implementing this approach shown here would have meant to have `Msg` + `ExternalMsg` + `CmdMsg` + `ExternalCmdMsg`. That had been a bit too much so I kept going with `Msg.Foo` and `Msg.CmdFoo`.)

TimLariviere:
>I also had one top level command to show alerts. Implementing this approach shown here would have meant to have Msg + ExternalMsg + CmdMsg + ExternalCmdMsg. 

@aspnetde That's interesting.
What are exactly your alerts? Are they alert dialogs or something else?

I'm trying to imagine when it would make sense to introduce the concept of `ExternalCmdMsg`.
Personally I would only use ExternalMsg to handle all external things. Like navigation or triggering commands on other pages.
This way there's no need to an `ExternalCmdMsg`. `CmdMsg` stays specific to its page.

```fsharp
type Msg =
    | SomeMsg

type ExternalMsg =
    | TriggerCmdOnOtherPage

type CmdMsg =
    | TriggerCmdOnThisPage

let update msg model =
    match msg with
    | SomeMsg -> model, (Some TriggerCmdOnOtherPage), TriggerCmdOnThisPage


/// The other page
type Msg =
    | PageATriggeredCmd

type CmdMsg =
    | TriggerCmdOnThisPage

let update msg model =
    match msg with
    | PageATriggeredCmd -> model, None, TriggerCmdOnThisPage
```

aspnetde:
For example I had a small function I didn't want to duplicate x times across the app. Like:

```fs
let private showErrorAlert errorMessage = async {
    do! Application.Current.MainPage
            .DisplayAlert("Error", errorMessage, "Okay") |> Async.AwaitTask
    return None
}
```

I guess that one could also be moved in some kind of helper module. But it's not hard to imagine a scenario where one would return a specific message type etcetera.

Anyway: No showstopper here!

src/Fabulous.Core/ViewHelpers.fs
```diff
@@ -92,6 +92,37 @@ module SimplerHelpers =
...
+                    | ValueSome pages -> pages
+                    | ValueNone ->
+                        match element.TryGetAttribute<ViewElement[]>("Children") with
+                        | ValueNone -> [||]
+                        | ValueSome children -> children
+
+            childElements
+            |> Seq.map (fun e -> e |> tryFindViewElement automationId)
+            |> Seq.filter (fun e -> e.IsSome)
+            |> Seq.map (fun e -> e.Value)
```
SergejDK: Is this mapping useful? You are only interested in the first element after all. So you could take the first and return the value.

TimLariviere: I'm doing this 2nd map here because the type returned by Seq.filter is `Seq<ViewElement option>`, so Seq.tryHead will return a `ViewElement option option`.

Without this map, I would need to match on the `ViewElement option option` and in the case `tryHead` found something, convert the `ViewElement option` to `ViewElement`.
I think it's faster to read and write the way I did.

Also the advantage of Seq here is that it won't evaluate the whole collection (like IEnumerable and LINQ in C#), it'll try each item one by one until it finds a matching one.

SergejDK: If it works like this then I didn't say anything. Was just curious about it because if it would work like LINQ it would be bad for performance.

cmeeren:
Excellent explanation in 922fa90. Might borrow from that (and the rest of this PR) for Elmish.WPF.

Might I suggest that you explicitly say that the `mkProgramWithCmdMsg` does not do anything "magic" and that the user would get the same result by writing trivial code to convert `update` and `init` themselves? IMHO that could help drive the point home that this is indeed just a general pattern where absolutely no magic (or even non-trivial) is involved on Fabulous' part.

cmeeren:
Wonderful! 👍

src/Fabulous.Core/ViewHelpers.fs
baronfel: You could use Seq.choose here instead of Seq.map and Seq.filter, which would auto-unwrap the option for you and reduce the option nesting.

PR #424, opened 2019-05-25
Fixes typo in CmdMsg docs

sasmithjr:
Hey there! I was reading through the new `CmdMsg` docs, and I noticed that the types didn't line up between what was defined in the DU (`TickTimer`) and what was being returned from `update` (`TimerTick`). This changes the DU to define `TimerTick`.


TimLariviere:
Thank you for fixing it!

PR #430, opened 2019-05-30
Add support for implicit converters when using Shell

TimLariviere:
Shell heavily relies on implicit operators to convert between the numerous types required to make a good hierarchy.

This PR adds some of them.

I will surely make other PRs when I encounter other cases


SergejDK:
@TimLariviere 
I think for the `Tab`-class and `TabBar` there should be implicit conversion, too. Normally they include ShellContent. They came in 4.0.

TimLariviere:
From what I saw, `Tab` and `TabBar` use the implicit conversion of ShellContent.
They don't have operators on their own.

SergejDK:
You are right! Just checked it... 
Well then I think it should be easier to write shell code now.

PR #420, opened 2019-05-23
Xamarin Essentials Update

SergejDK:
fix #419

PR #434, opened 2019-06-02
WIP: Migrate to NuGet package reference

mallibone:
Migrate iOS and Android templates to package reference. My main reason for doing this was having quite a few problems updating NuGet packages in the IDEs with the old `package.config` format.

NuGet Package updates:
* JSON.Net 12.0.2
* [.Net Core 4.6.2](https://devblogs.microsoft.com/dotnet/announcing-f-4-6/)
* [Android Support Libs to 28.0.0.1](https://devblogs.microsoft.com/xamarin/multi-targeting-with-android-support-library-v28/)
* Xamarin Essentials 1.1.0

Further housecleaning:
* Android Target SDK28 / Version 9.0


SergejDK:
@mallibone 
for the update of Xamarin.Essentials please take a look into: #420 
There @TimLariviere  and me decided to stay with 1.0.0

TimLariviere:
Do iOS and Android fsprojs support `PackageReference` now?

I opened an issue some times ago because it wasn't fully supported on VS Windows. (no restore, no tooling support)
The answer was that no further work would be done on the F# legacy project system.
And the 2 options were:
- Stay on old fsproj with packages.config
- Move to new fsproj format (though Xamarin does not support it yet)

See https://github.com/dotnet/fsharp/issues/5849 and https://github.com/dotnet/fsharp/pull/5853

mallibone:
So let's just say I got ahead of myself and will no longer only take my MacBook for developing something like this 🙈 - I will close the pull request... What initially brought me down this road was that I had to update various NuGet packages for enabling the Material design.

There is anyway still something off - the test build has an issue. Opening the same project with VS (on a Mac) compiles just fine. But I guess that is a future problem when Package Reference is supported on Windows 🙁

TimLariviere:
@mallibone It's not necessarily bad. It will be eventually the case someday (let's hope sooner than later 😄).

Maybe we can accept this as an optional flag to `dotnet new`?
Like `dotnet new fabulous-app --PackageReference -n MyApp`

mallibone:
@TimLariviere yeah, it would be nice and also greatly ease the upgrade to the Android Support Libs v28 / Android 9 mentioned in #442 - but there is another issue with the tests I first have to find. 🤔

But will check into the optional reference - and who knows one day one might add different templates when creating a new fabulous project (tabs, multi-page etc.)

Issue #442, opened 2019-06-09
Update Android Support Libraries to v28 (matching Android 9.0)


TimLariviere:
Like said by @PureWeen, we should update the templates to use the v28 packages of the Android Support Libraries.

Also I'm reconsidering the rule of the least major version for the templates #419
It's a good rule for our packages, since it would otherwise prevent users from using an earlier version of an external package we use if they really need it (like Newtonsoft.Json)
But is it really applicable to the templates?

Users are not blocked by it, but need to update the packages (like Xamarin.Essentials) each time they create a new template.

@SergejDK What do you think?

mallibone:
Just to chime in - updating the libs to v28 and Android 9.0 would also allow the usage of Material Visuals aka beautiful Material designed apps.

As of the versions of NuGet packages (just my opinion 😇): Since there is no long term supported version of Xamarin Forms. I believe keeping it aligned with the C# template in VS could be a good strategy for now.

In general, I would concur with using the latest and greatest in the templates, i.e. not letting the user having to update to all the latest libs but keeping the dependencies from Fabulous as low as possible.

If it is not to time pressing or you are not already all over it, I could tackle the update. There are some different packages required for 9.0 as are currently referenced and the target files do not seem to update when updating the packages via VS2019 NuGet package manager - any idea why that may be?

TimLariviere:
>the target files do not seem to update when updating the packages via VS2019 NuGet package manager - any idea why that may be?

@mallibone What files do you mean?

For the samples, the Android packages are downloaded by Paket but the path to the dlls are set by hand. 
For the templates, the packages are managed by NuGet (`packages.config`) but the version is replaced by a token declared in https://github.com/fsprojects/Fabulous/blob/master/templates/content/blank/.template.config/template.json

Maybe Visual Studio doesn't support updating `packages.config` when using tokens instead of versions?

>If it is not to time pressing or you are not already all over it, I could tackle the update

Sure. You're more than welcome to do it. Thanks! :)

SergejDK:
The update to v28 is a great idea - I think there are some people who want to use Visuals.

> Users are not blocked by it, but need to update the packages (like Xamarin.Essentials) each time they create a new template.

If you create a new project you will normally have the latest packages in it or you want them in it because you have no dependencies in a new project to older parts. I do not know one person who do not update when they start a new project.
We should say 'updates only to stable versions' - just to be clear.

[...]

PR #446, opened 2019-06-10
Upgrade samples to Android Pie (9.0)

TimLariviere:
Partial cover of #442

Update the samples to compile for Android Pie (9.0) instead of Oreo (8.1).
Also samples now start support at Android Lollipop 5.0 (API Level 21) instead of Android KitKat 4.4

This is required to be able to use Shell and Visual.


SergejDK:
2 things: 

1. Maybe we should write sth. in the docs that the `.fake` directory should be deleted else it won't work when running `./build.sh` - it will fail at `BuildFabulous`.

2. The Calculator App does not work anymore. Can't make it work on my device (Motorola G6 Plus).

TimLariviere:
>Works now. Configurationmanager needed to provide Calculator.Droid (wasn't set as porvided like the other projects); rebuild and restart after this worked.

@SergejDK I'm not sure to understand.
What do you mean by "ConfigurationManager needed to provide Calculator.Droid"?

SergejDK:
@TimLariviere have it on german so i hope the follwoing screenshot shows what i mean:
grafik
grafik

SergejDK:
the right column i think translated should be "provided" or "supplied"...
If you check this it works.

TimLariviere:
Ah ok. In English, the columns are "Build" and "Deploy".
So makes sense, it weren't working.

Visual Studio prevented the sample to be deployed on your device.

PR #449, opened 2019-06-11
Fix Android samples not deploying in some configurations

TimLariviere:
Some samples were not able to deploy on specific configurations (Debug/Release, AnyCPU/iPhone/etc.)

This is now fixed. All samples build and deploy the same.

@SergejDK This should fix the issue you had


SergejDK:
Looks good! 
Works! - Thanks!

Both Columns for UWP are empty - is this right?

TimLariviere:
UWP can only be built and deployed for x86 and x64
It does not support AnyCPU (and derived ARM/iPhone configurations)

SergejDK:
Okay, great. Didn't know exactly about it.
Thanks :)

SergejDK:
Thanks, again!

PR #447, opened 2019-06-10
Add missing properties to Shell related controls

TimLariviere:
This PR adds properties to the bindings file that were missing.
Also fixes some converters.

Note: Some changes to the bindings file feel hacky.
Really would like to rework that later.

PR #450, opened 2019-06-12
Add the Fabimals sample

TimLariviere:
Now that #446 and #447 are merged, the Fabimals sample can finally be added to Fabulous.

It is a direct port of the official Xaminals sample, to demonstrate how to use Shell in Fabulous.
https://github.com/xamarin/xamarin-forms-samples/blob/master/UserInterface/Xaminals

I feel like the implementation is ok, except for Routing which is currently being reworked in Xamarin.Forms (https://github.com/xamarin/Xamarin.Forms/issues/5166)
But don't hesitate to suggest ways I can improve this sample.


SergejDK:
@TimLariviere 
is this the last PR of the big one?

TimLariviere:
Yes it is.

TimLariviere:
The code size is roughly 750 lines of F# (without accounting for the Data folder).

SergejDK:
@TimLariviere 
could you maybe delete the `Shell`-Part in the AllControls because it is not needed anylonger cause auf your great example!
(Currently checking)

> The code size is roughly 750 lines of F# (without accounting for the Data folder).

Not too bad.

SergejDK:
@TimLariviere 
tested it so far - didn't look at the code yet! It works, thats great! 

It is kind of laggy on my device in debug mode but I think this was discussed a lot in the issues - "Performance while navigating" etc. so we have this covered there.

I needed to reference the `fabimals`project in `Fabimals.Droid`project again - so it looks like this for me now: 
grafik

samples/Shell/Fabimals/Droid/MainActivity.fs
```diff
@@ -0,0 +1,24 @@
...
+open Android.Views
+open Android.Widget
+open Android.OS
+open Xamarin.Forms.Platform.Android
+
+[<Activity(Label = "Fabimals", Icon = "@mipmap/icon", Theme = "@style/MainTheme", MainLauncher = true, ConfigurationChanges = (ConfigChanges.ScreenSize ||| ConfigChanges.Orientation))>]
+type MainActivity() =
+    inherit FormsAppCompatActivity()
+    override this.OnCreate(bundle : Bundle) =
+        FormsAppCompatActivity.TabLayoutResource <- Resources.Layout.Tabbar
```
SergejDK: We should point out that Toolbar/TabLayoutResource is needed.

samples/Shell/Fabimals/Fabimals/App.fs
```diff
@@ -0,0 +1,271 @@
...
+open Fabimals.Data
+open Fabimals.Models
+open Fabimals.Components
+open Fabimals.Routes
+
+module AppStyles =
+    let applyDefaultShellItemStyle (element: ViewElement) =
+        element.ShellBackgroundColor(Color.FromHex("#455A64"))
+               .ShellForegroundColor(Color.White)
+               .ShellTitleColor(Color.White)
```
SergejDK: You define `ShellTitleColor` 2 times.
.ShellTitleColor(Color.White) and .ShellTitleColor(Color.FromHex("#B4FFFFFF"))

I think one time should be enough?

samples/Shell/Fabimals/Fabimals/Components/AnimalListPage.fs
```diff
@@ -0,0 +1,63 @@
...
+    let navigateToAfterSelectionChanged dispatch (args: SelectionChangedEventArgs) =
+        match args.CurrentSelection |> Seq.tryHead with
+        | None -> ()
+        | Some item ->
+            let data = item :?> ItemListElementData
+            let animal = data.Key.GetAttributeKeyed(ViewAttributes.TagAttribKey) :?> Animal
+            dispatch (SelectAnimal animal)
+
+    let view model dispatch =
+        dependsOn (model.PageTitle, model.IsTopBarDisplayed, model.AllAnimals, model.FilteredAnimals) (fun model (pageTitle, isTopBarDisplayed, allAnimals, filteredAnimals) ->
```
SergejDK: you could replace `model` with an underscore. model is not used in there.

samples/Shell/Fabimals/Fabimals/Components/AnimalDetailsPage.fs
```diff
@@ -0,0 +1,60 @@
...
+open Fabulous.DynamicViews
+open Xamarin.Forms
+open Fabimals.Models
+
+module AnimalDetails =
+    type Msg = NoOp
+    type CmdMsg = NoOp
+    type ExternalMsg = NoOp
+    
+    type Model =
```
SergejDK: The Model is the `Animal` Model. 
Couldn't it be reused?

samples/Shell/Fabimals/Fabimals/Views/About.fs
```diff
@@ -0,0 +1,144 @@
...
+
+    let view dispatch =
+        dependsOn () (fun _ () ->
+            View.ContentPage(
+                title="About",
+                content=View.ScrollView(
+                    View.Grid(
+                        rowdefs=["auto"; "*"],
+                        children=[
+                            View.StackLayout(
```
SergejDK: Don't know it for sure but nested Stacklayouts are always heavy in performance. I think Flex or Grid could be better in it?
Generally I think the About page is kinda interesting. You have a Scrollview with a Grid and a scrollview again but I think there is no need for this.

Scrollview and a Grid should be enough?

samples/Shell/Fabimals/Fabimals/App.fs
TimLariviere: Good catch! The 2nd one is a failed copy-paste.

samples/Shell/Fabimals/Fabimals/Components/AnimalListPage.fs
TimLariviere: Here `model` is of type `DoNotUseModelInsideDependsOn`
It needs to match the name `model` to prevent access to the real model inside `dependsOn .
https://github.com/fsprojects/Fabulous/blob/dcf5dbe175fa617977a6aa08a5a25dff25f5c292/src/Fabulous.Core/ViewHelpers.fs#L24-L30

samples/Shell/Fabimals/Fabimals/Components/AnimalDetailsPage.fs
TimLariviere: I think it's a best practice to keep it separated, even if it means duplicating the type.
`Animal` is a data type, `Model` is a state type.

samples/Shell/Fabimals/Fabimals/Views/About.fs
TimLariviere: You're right. I'll take a closer look at this page to see what can be simplified.
I only translated the AboutPage.xaml from Xaminals without asking too much questions. :)

samples/Shell/Fabimals/Droid/MainActivity.fs
TimLariviere: Is it?
That's interesting. Maybe that's why the Android app wasn't working at first.
I'll try without, see what kinds of issue it raises.

samples/Shell/Fabimals/Fabimals/Components/AnimalDetailsPage.fs
TimLariviere: That said, maybe changing `Model` to contain an `Animal` would be better than duplicating all the fields?

samples/Shell/Fabimals/Droid/MainActivity.fs
SergejDK: AFAIK yes but would be greta if you could test this.

samples/Shell/Fabimals/Fabimals/Components/AnimalListPage.fs
SergejDK: Alright, didn't know about that.

samples/Shell/Fabimals/Fabimals/Components/AnimalDetailsPage.fs
SergejDK: > That said, maybe changing Model to contain an Animal would be better than duplicating all the fields?

That would be better, yes. I think this way it is a bit cleaner.

samples/Shell/Fabimals/Fabimals/Views/About.fs
SergejDK: Maybe this could be interesting: [https://github.com/jsuarezruiz/xamarin-forms-perf-playground](https://github.com/jsuarezruiz/xamarin-forms-perf-playground)

There are some explanation on do's and dont's

samples/Shell/Fabimals/Fabimals/App.fs
TimLariviere: This is fixed now

samples/Shell/Fabimals/Fabimals/Components/AnimalDetailsPage.fs
TimLariviere: Fixed too

TimLariviere:
>I needed to reference the fabimalsproject in Fabimals.Droidproject again - so it looks like this for me now

Should be fixed now

samples/Shell/Fabimals/Droid/MainActivity.fs
TimLariviere: Doesn't seem to be a requirement. Just tried with those 2 lines commented out and Fabimals just works the same.

TimLariviere:
Should be noted that using the search bar on a real iPhone currently crashes.
Opened an issue for that: https://github.com/xamarin/Xamarin.Forms/issues/6505

samples/Shell/Fabimals/Fabimals/Components/SearchHandlers.fs
```diff
@@ -0,0 +1,42 @@
...
+module SearchHandlers =
+    type Msg =
+        | QueryChanged of string
+        | AnimalSelected of Animal
+
+    let animalSearchHandler animals dispatch =
+        View.SearchHandler(
+            placeholder="Enter search term",
+            showsResults=true,
+            queryChanged=(fun (_, newValue) -> dispatch (QueryChanged newValue)),
```
SergejDK: Do you think we can debounce this? 
I mean that not on every change the message is run - just curious if it is possible.

SergejDK:
@TimLariviere 
look through the code and it looks good to me.

samples/Shell/Fabimals/Fabimals/Components/SearchHandlers.fs
TimLariviere: It is possible, yes. It's quite easy:
```fsharp
queryChanged=debounce 250 (fun (_, newValue) -> dispatch (QueryChanged newValue)),
```

I tried it with different times, but the UX doesn't feel satisfying.
The delay is the UI is not great.

SergejDK: hmmm... does every message run or only the latest with the debounce ?
maybe if only the latest massage will run it could be a bit better.

SergejDK:
Saw this on stream.
Bildschirmfoto 2019-06-14 um 19 35 43

The first item looks kinda broken. Didn't saw when it broke - I think after the search?

TimLariviere:
@SergejDK Yes, I did open an issue on the XF repo for that.
https://github.com/xamarin/Xamarin.Forms/issues/6507

When you leave the page with a CollectionView and come back to it, for some reasons, the previous items are still there stuck at the top.

SergejDK:
@TimLariviere 
good thing!

TimLariviere:
I want to try to make it work with ShellContent.ContentTemplate instead of Content before merging it.

SergejDK:
Do you mean it as we do for listview?

samples/Shell/Fabimals/Fabimals/Components/SearchHandlers.fs
TimLariviere: debounce only sends the last message in a given time frame.
Everytime the event is triggered, it will wait for 250ms to check if there's no other event triggers. Only then it really calls the function and dispatches.

So this delay is kinda bad when typing. It can give a few quirks like if you're typing too slow, it will show delayed results not matching what you're currently typing.
E.g. You're typing "ABCD" slowly. While you wrote "ABC", it will display results for "AB".

So it's confusing.

TimLariviere:
Yes. Just like we discussed it on the stream. Through the use of IShellContentController.
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Shell/IShellContentController.cs

samples/Shell/Fabimals/Fabimals/Components/SearchHandlers.fs
SergejDK: well i think maybe playing a bit with the timeframe could be a good way. I mean RX or ReactiveUI uses nearly the same logic for throttle or sth. else. There it works great, so I thought this could be a great thing to have. 

Maybe we need to investigate on this and make mit more usable -> new issue?

TimLariviere:
https://github.com/fsprojects/Fabulous/blob/b13e7c66f27a0ef754db7afb19afca3dacc9fdcb/src/Fabulous.Core/ViewConverters.fs#L203-L206

@PureWeen Would you know a better way to achieve this?
`Xamarin.Forms.Internals.IDataTemplate` is marked obsolete, but it seems to be the only way to change how the view should be constructed by a DataTemplate.

PR #451, opened 2019-06-12
I got bit by this...

Happypig375:


docs/views-skiasharp.md
```diff
@@ -39,6 +39,8 @@ View.SKCanvasView(enableTouchEvents = true,
 )
 ```
 
+By default, the view will not be redrawn when the model changes. You should call `View.SKCanvasView(..., invalidate = true)` if you want the drawn content to depend on model changes.
```
TimLariviere: Rewording it a bit
````suggestion
By default, the view will not be redrawn when the model changes. You should set `invalidate` to true when you know that a redraw is needed. Set it back to false when done, otherwise it will be redrawn at each update.
```fsharp
View.SKCanvasView(..., invalidate = true)
```
````

TimLariviere:
Thanks! 

We forgot to add this precision in the docs when we added `invalidate`.
Sorry about that.

PR #476, opened 2019-06-26
added example images

CMThre3:
example images have been added to (most of) the view elements, some do not have images as they cannot be demonstrated with images such as touch gestures


dsyme:
I'm pulling this as Kyle is going to work on splitting up the `view-elements` doc page into a few bits next, I don't want to conflict this

dsyme:
I looked at the images and they do look good.

PR #482, opened 2019-06-27
paket update

dsyme:
This is the result of running paket update in the repository

@TimLariviere I don't actually see any changes here we want and they may actually be wrong.  The Paket.Restore.targets already gets imported in `build.props` for example


dsyme:
I'll close this, as I don't think we want it, but leaving it here for visibility

TimLariviere:
@forki Fabulous uses Paket, but not in a conventional way.
Paket is only used as a CLI tool to download packages for the samples. It does not manage the references inside fsproj files.

@dsyme Yes, it's a bit annoying. Paket tries to manage projects that aren't managed by it.
`paket update` will make all these changes each time, but we don't want them.

I don't know if there is an option on Paket to tell it to only update the lock file and don't care about the rest.

The solution I found was to create an empty `paket.dependencies` inside the folder we don't want Paket to go in.
https://github.com/fsprojects/Fabulous/blob/master/templates/paket.dependencies

We can generalize that and put those `paket.dependencies` file in every folder like `src`, `extensions`, etc.?

PR #490, opened 2019-07-02
Update README to match the new structure

TimLariviere:
Tried to reflect and explain the new structure in the README file.

I fear it might not be "simply explained" and/or compelling enough for people coming on github.com/fsprojects/Fabulous


README.md
```diff
@@ -1,11 +1,49 @@
...
+
+- Generator
+
+An external tool that will generate Fabulous-compatible wrappers to enable using existing controls (or components libraries) with a dynamic UI code. This tool is configured with a bindings json file.  
+Fabulous.XamarinForms uses it to generate all of its View DSL.
+
+- LiveUpdate / fabulous-cli
+
+A combo of a dotnet CLI tool and a plugin for Fabulous. They enable your users to write their apps and see live their changes in the running emulator / real device without restarting.  
+It is similar to the Hot Reload module of Flutter.
```
SergejDK: maybe hint out that it is still in early development?

README.md
```diff
@@ -1,11 +1,49 @@
...
+
+### Fabulous for Xamarin.Forms
+
+This library allows you to use the ultra-simple Model-View-Update architecture to build applications for iOS, Android, Mac, WPF and more using Xamarin.Forms. It is built on Fabulous.
+
+### Fabulous StaticView
+
+This library allows you to write Xamarin.Forms apps using XAML and the Model-View-Update architecture.
```
SergejDK: We have this but do we activly maintain it ? I mean maybe point this out if we do not.

README.md
TimLariviere: You're right! Updated.

README.md
TimLariviere: Same here. Updated :)

TimLariviere:
Changed the README based on the feedback I received from some C# people.
The README has been split in different files :
- Primary README is only a brief explanation of what we can find on this repo
- Each part has its own README (Fabulous has a FABULOUS.md instead)
- Roadmap has been split and integrated inside its corresponding README

PR #493, opened 2019-07-04
Upgrade Fabulous.XamarinForms to XF 4.1

### LINKED ISSUE #426: Collection/Carousel-View, Shell
Not every of this control is working currently correctly. 
Collection works fine for iOS but not Android.
Carousel seems to work on Android but not iOS.

Shell needs a Test.

It would be great if we could make this work.

SergejDK:
With the upgrade to 4.1 the following issues are fixed:

fix #426 

The fuget-link: [https://www.fuget.org/packages/Xamarin.Forms/4.1.0.555618/lib/netstandard2.0/diff/4.0.0.497661/](https://www.fuget.org/packages/Xamarin.Forms/4.1.0.555618/lib/netstandard2.0/diff/4.0.0.497661/)

### DIFF
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -20,16 +20,15 @@ jobs:
     vmImage: 'macOS-10.14'
   steps:
   - bash: |
-          SYMLINK=5_18_1
-          MONOPREFIX=/Library/Frameworks/Mono.framework/Versions/$SYMLINK
+          MONOPREFIX=/Library/Frameworks/Mono.framework/Versions/Current/Commands
           echo "##vso[task.setvariable variable=DYLD_FALLBACK_LIBRARY_PATH;]$MONOPREFIX/lib:/lib:/usr/lib:$DYLD_LIBRARY_FALLBACK_PATH"
           echo "##vso[task.setvariable variable=PKG_CONFIG_PATH;]$MONOPREFIX/lib/pkgconfig:$MONOPREFIX/share/pkgconfig:$PKG_CONFIG_PATH"
           echo "##vso[task.setvariable variable=PATH;]$MONOPREFIX/bin:$PATH"
     displayName: Set Mono 
   - bash: 'mono --version'
     displayName: 'Mono version'
-  - bash: '$PATH'
-    displayName: 'Show Path variable'
+  - bash: 'echo $PATH'
+    displayName: 'Showing Pathvarialbe'
   - script: ./build.sh Test
   - task: PublishTestResults@2
     inputs:
--- a/azure-pipelines-pr.yml
+++ b/azure-pipelines-pr.yml
@@ -24,7 +24,7 @@ jobs:
           MONOPREFIX=/Library/Frameworks/Mono.framework/Versions/$SYMLINK
           echo "##vso[task.setvariable variable=DYLD_FALLBACK_LIBRARY_PATH;]$MONOPREFIX/lib:/lib:/usr/lib:$DYLD_LIBRARY_FALLBACK_PATH"
           echo "##vso[task.setvariable variable=PKG_CONFIG_PATH;]$MONOPREFIX/lib/pkgconfig:$MONOPREFIX/share/pkgconfig:$PKG_CONFIG_PATH"
-          PATH=/Library/Frameworks/Mono.framework/Versions/Current/Commands/bin:/usr/local/lib/ruby/gems/2.6.0/bin:/usr/local/opt/ruby/bin:/usr/local/opt/curl/bin:/usr/local/bin:/usr/local/sbin:/Users/vsts/bin:/Users/vsts/.yarn/bin:/usr/local/go/bin:/Users/vsts/Library/Android/sdk/tools:/Users/vsts/Library/Android/sdk/platform-tools:/Users/vsts/Library/Android/sdk/ndk-bundle:/usr/bin:/bin:/usr/sbin:/sbin:/Users/vsts/.azcopy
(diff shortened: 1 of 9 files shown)


SergejDK:
@TimLariviere 
do you know sth. about this error? 
it is kinda strange - locally it works... Thought of different mono version maybe but kinda seems not like it.

TimLariviere:
The real error seems to be
> Mono.Cecil.ResolutionException: Failed to resolve AppKit.NSMenuItem[] AppKit.NSMenu::get_Items()

SergejDK:
So I think this is a XF issue then ?

TimLariviere:
Yes, I think we can open an issue for that and include the stack trace error.
We're using the latest version of Xcode, so there shouldn't be an exception.

SergejDK:
Is this reproducable locally on your machine? Or even maybe with a whole ne project XF 4.1 and C#.
Would like to test this instead of opening an issue right now.  (Currently at work and only have windows here - if you don't have the time I'll be testing it later today)

If it can't be reproduced maybe there is another issue on the Azure Pipeline side? 
Else we have a reproduction for the XF Team.

TimLariviere:
>Is this reproducable locally on your machine?
No, I do not reproduce it

SergejDK:
XF with F#, C# locally are working fine, too....
This is kinda interesting.

Will open an issue but this is maybe a config problem in Azure Pipepline or our side .

TimLariviere:
All's great. :)
Except for the SelectedItems property.

Do you think after that we should merge even if CI is breaking?

I'm planning to release a version 0.40 today or tomorrow

TimLariviere:
(Just removed the verbosity)

TimLariviere:
Regarding our error with Menu on macOS Release mode, https://github.com/xamarin/Xamarin.Forms/pull/4718 has been merged in XF 4.1.0.
So it might be the reason why it fails now.

SergejDK:
> Except for the SelectedItems property.

Will check it directly and make a fix for this in the next hour.

> Regarding our error with Menu on macOS Release mode, xamarin/Xamarin.Forms#4718 has been merged in XF 4.1.0. So it might be the reason why it fails now.

Saw the PR and it seems like this is the problem because nothing else was moved into 4.1 so far.
If we merged this we should be aware of this issue and that maybe won't find any other issues. Normally I would say YES because of Collection/CarouselView but srsly it could be that we break more things. So because of this danger I would NOT merge it.

SergejDK:
> Missing update: we can no longer set SelectableItemsView's SelectedItems property.
We need to remove it from the binding (with its function if not used elsewhere)
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/release-notes/4.1/4.1.0-api#type-changed-xamarinformsselectableitemsview

Just checked it but we only set it through `updateCollectionGeneric`. So I think this way it is okay?

TimLariviere:
I would have said it depends if the list we get is read only or not.

But after reading the source code, it appears the property hasn't changed in XF 4.1.0.555618.
https://github.com/xamarin/Xamarin.Forms/blob/release-4.1.0-sr1/Xamarin.Forms.Core/Items/SelectableItemsView.cs#L44-L48

Even after testing it in VS, the setter is still there.
So you can ignore my comment. :)

SergejDK:
For Follow-Up

https://github.com/xamarin/Xamarin.Forms/issues/6796

SergejDK:
@TimLariviere 
just checked the azure-pipelines set-up of XF.
```
- name: MONO_VERSION
  value: 5_18_1
- name: XCODE_VERSION
  value: 10.2
- name: NUGET_VERSION
value: 5.0.2
```
I think the only difference would be the Mono-versions?
Should we try to use 5.18 instead of 5.12? 
I can't imagine that XF pipelines does not show this error....

TimLariviere:
Oh, we weren't using the latest version of Mono?
Yes, if you could test to change that and see if it fixes our issue.

That would be great if it's only that. :)

SergejDK:
Never worked with pipelines.. Thought setting the Variables in yml should make it work but kinda not.
Do you know it?

SergejDK:
Setting in Azure pipelines does not make monopath update for fake.
We need to somehow set the path while running FAKE. That should do the magic

SergejDK:
@TimLariviere 
the last one was using 5.18.1 but did not work. Same issue....

SergejDK:
Looks good so far(AllControls works) - so I hope it will work!

azure-pipelines-pr.yml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+    displayName: Switch to the latest Xamarin SDK
+  - bash: echo '##vso[task.setvariable variable=MD_APPLE_SDK_ROOT;]'/Applications/Xcode_10.2.app;sudo xcode-select --switch /Applications/Xcode_10.2.app/Contents/Developer
+    displayName: Switch to the latest Xcode
+  - bash: |
+          SYMLINK=5_18_1
+          MONOPREFIX=/Library/Frameworks/Mono.framework/Versions/$SYMLINK
+          echo "##vso[task.setvariable variable=DYLD_FALLBACK_LIBRARY_PATH;]$MONOPREFIX/lib:/lib:/usr/lib:$DYLD_LIBRARY_FALLBACK_PATH"
+          echo "##vso[task.setvariable variable=PKG_CONFIG_PATH;]$MONOPREFIX/lib/pkgconfig:$MONOPREFIX/share/pkgconfig:$PKG_CONFIG_PATH"
+          export $PATH=/Library/Frameworks/Mono.framework/Versions/5.18.1/Commands/bin:/usr/local/lib/ruby/gems/2.6.0/bin:/usr/local/opt/ruby/bin:/usr/local/opt/curl/bin:/usr/local/bin:/usr/local/sbin:/Users/vsts/bin:/Users/vsts/.yarn/bin:/usr/local/go/bin:/Users/vsts/Library/Android/sdk/tools:/Users/vsts/Library/Android/sdk/platform-tools:/Users/vsts/Library/Android/sdk/ndk-bundle:/Library/Frameworks/Mono.framework/Versions/Current/Commands:/usr/bin:/bin:/usr/sbin:/sbin:/Users/vsts/.azcopy
```
SergejDK: this is kinda bad....
The idea here is to remove the current Path for mono.
Then only add the new 15.8.1 so FAKE will use this.
Didn't knew how to remove something from the path so that was a "quick&dirty" solution.
Need to cleanup this stuff.

SergejDK:
Creating a new clearer PR.

PR #494, opened 2019-07-05
Remove invalid targets + move import targets before dependencies

TimLariviere:
Fix issues reported in https://github.com/fsprojects/Fabulous/issues/385#issuecomment-507554245 and https://github.com/fsprojects/Fabulous/issues/385#issuecomment-507576046


TimLariviere:
Thanks!
Merging...

Issue #499, opened 2019-07-06
Setting useSafeArea to false has no effect


JamesRandall:
As per the title setting this property has no effect. I've got this Fabulous version of the boilerplate Xaml:

```
let view (model: Application.Model) dispatch =
        View.ContentPage(useSafeArea=false,
                         content=View.FlexLayout(
                                                    direction=FlexDirection.Row,
                                                    justifyContent=FlexJustify.End,
                                                    children=[
                                                        View.Label(text="Welcome to Xamarin.Forms!", horizontalOptions=LayoutOptions.Center, verticalOptions=LayoutOptions.CenterAndExpand)
                                                    ]
                                                ))
```

XAML equivalent:

```
<?xml version="1.0" encoding="utf-8"?>
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" mc:Ignorable="d" x:Class="safeareatest.MainPage"
             xmlns:ios="clr-namespace:Xamarin.Forms.PlatformConfiguration.iOSSpecific;assembly=Xamarin.Forms.Core"
             ios:Page.UseSafeArea="false">
    <FlexLayout Direction="Row" JustifyContent="End">
        <!-- Place new controls here -->
        <Label Text="Welcome to Xamarin.Forms!" HorizontalOptions="Center" VerticalOptions="CenterAndExpand" />
    </FlexLayout>
</ContentPage>

```
The latter places the label under the status bar while the former (F#) places it under the status bar.

I'm still finding my way around both Fabulous and F# but it looks to me like an issue in how the property is set here:

https://github.com/fsprojects/Fabulous/blob/e1613ad8a1cd7917f871e945fd21049ab8da87f8/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json#L1631

If I'm reading that write it's always setting the property to true regardless of what is set in the view code. Fabulous seems to default to the safe area being used and so you can never turn it off.

(I've got Fabulous building so if I get a chance I'll submit a PR - just a bit new to all this)

SergejDK:
Yes you are right. It is always true and does it does not care about setting it.

I don't know the specific reason why it was done this way but I think we should do it as XF does it.
[...]

PR #500, opened 2019-07-06
Added a view converter that sets the useSafeArea property on a page

JamesRandall:
Proposed fix for issue #499 

I've tested it setting to true, setting to false and not set using the AllControls.fs project. Thanks for your help.

### DIFF
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
@@ -1628,7 +1628,7 @@
           "name": "UseSafeArea",
           "inputType": "bool",
           "modelType": "bool",
-          "updateCode": "(fun _ _ target -> Xamarin.Forms.PlatformConfiguration.iOSSpecific.Page.SetUseSafeArea((target : Xamarin.Forms.Page).On<Xamarin.Forms.PlatformConfiguration.iOS>(), true) |> ignore)"
+          "updateCode": "updateUseSafeArea"
         },
         {
           "name": "Appearing",
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
@@ -1628,6 +1628,7 @@
           "name": "UseSafeArea",
           "inputType": "bool",
           "modelType": "bool",
+          "defaultValue": "false",
           "updateCode": "updateUseSafeArea"
         },
         {
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms/ViewConverters.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms/ViewConverters.fs
@@ -817,3 +817,10 @@ module Converters =
             let realTarget = (target :> Xamarin.Forms.IShellContentController).Page
             if realTarget <> null then currValue.UpdateIncremental(prevValue, realTarget)            
         | ValueSome _, ValueNone -> target.ContentTemplate <- null
+        
+    let updateUseSafeArea _ (currValue: bool voption) (target: Xamarin.Forms.Page) =
+        match currValue with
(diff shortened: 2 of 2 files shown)


Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
```diff
@@ -1628,7 +1628,7 @@
           "name": "UseSafeArea",
           "inputType": "bool",
           "modelType": "bool",
-          "updateCode": "(fun _ _ target -> Xamarin.Forms.PlatformConfiguration.iOSSpecific.Page.SetUseSafeArea((target : Xamarin.Forms.Page).On<Xamarin.Forms.PlatformConfiguration.iOS>(), true) |> ignore)"
+          "updateCode": "updateUseSafeArea"
```
SergejDK: It's good that you used ViewConverter for this!
One thing is missing: `"defaultValue": "false"`. without this the default would be true and this is the wrong default as XF defines it.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/ViewConverters.fs
```diff
@@ -816,4 +816,11 @@ module Converters =
             target.ContentTemplate <- ViewElementDataTemplate(currValue)
             let realTarget = (target :> Xamarin.Forms.IShellContentController).Page
             if realTarget <> null then currValue.UpdateIncremental(prevValue, realTarget)            
-        | ValueSome _, ValueNone -> target.ContentTemplate <- null
\ No newline at end of file
+        | ValueSome _, ValueNone -> target.ContentTemplate <- null
+        
+    let updateUseSafeArea _ (currValue: bool voption) (target: Xamarin.Forms.Page) =
+        match currValue with
```
SergejDK: the previous Value should be used, too. You can check the other updateCodes on the behavior.

this is just so fabulous won't call functions if they don't need to be called. Srsly I am not good at explaining .....

TimLariviere: Don't beat yourself up 😄 
It was understandable, at least for me. :)

To be efficient, Fabulous.XamarinForms uses some view diffing techniques to avoid calling Xamarin.Forms for nothing.
And comparing the previous value vs the current one help call Xamarin.Forms only when the value really changed.
That's why you can see other functions following the same pattern of matching both prevValue and currValue.

JamesRandall: It made sense to me! I'll make the change.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
JamesRandall: I was a little concerned of setting the default value as it seemed to be set to true in my 0.36 based project (given the issue was I couldn't turn it off) and if I set it to false that would change the layouts of potentially a lot of forms people had already written when they upgraded.

SergejDK: Well you are right and it was good that you were concerned but cause we are in a low version this is something that the devs should be aware of. 
If this will happen release 1.0 and higher we should definetly think about this and hint them to sth. like "breaking changes".

TimLariviere: Starting v1.0, any breaking change will wait for a v2.0 release

>as it seemed to be set to true in my 0.36 based project (given the issue was I couldn't turn it off)
The value was defaulting to true because of this:
https://github.com/fsprojects/Fabulous/blob/bfd59882bcf4a4877cdd62062269a48d6bd4c3e7/Fabulous.XamarinForms/src/Fabulous.XamarinForms/CustomControls.fs#L210-L213

It's still the case, it will still be true by default when we use `View.ContentPage`.
In the json file, the default value is only applied when we remove an attribute we were declaring previously
`View.ContentPage(useSafeArea=true)` -> `View.ContentPage()` = the default value (false) will be applied to "undo" what the use of `useSafeArea` did.

PR #511, opened 2019-07-08
XF Update 4.1

SergejDK:
This PR replaces the #493

Azure pipeline build works
A simple fix on build.fsx
Updating XF


Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
```diff
@@ -1628,8 +1628,7 @@
           "name": "UseSafeArea",
           "inputType": "bool",
           "modelType": "bool",
-          "defaultValue": "false",
-          "updateCode": "updateUseSafeArea"
+          "updateCode": "(fun _ _ target -> Xamarin.Forms.PlatformConfiguration.iOSSpecific.Page.SetUseSafeArea((target : Xamarin.Forms.Page).On<Xamarin.Forms.PlatformConfiguration.iOS>(), true) |> ignore)"
```
SergejDK: This part here.

azure-pipelines-full.yml
```diff
@@ -30,6 +36,16 @@ jobs:
   pool:
     vmImage: 'macOS-10.14'
   steps:
+  - bash: sudo $AGENT_HOMEDIRECTORY/scripts/select-xamarin-sdk.sh $(MONO_VERSION)
```
TimLariviere: Is it possible to extract those scripts in an other file somehow to avoid repeating them in both yml files?

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Controls/Xamarin.Forms.Core.json
TimLariviere: Could you revert this change, please? :)

azure-pipelines-full.yml
SergejDK: Yes we can have a 'default-pipeline' where this is set. Saw this in some other examples.

SergejDK:
> If you know of a way we could automate these files, it would be awesome.

Will update them manually this time but I'll see what is possible. Maybe just some Regex-magic.

> It was really just an issue with Mono/Xamarin versions then?

Yes. Mono, Xamarin, XCode. We should be fine after this :)

SergejDK:
> <dependencies>
      <dependency id="Fabulous.XamarinForms" version="0.40.4" />
      <dependency id="SkiaSharp.Views.Forms" version="1.60.3" />
    </dependencies>

Normally I do not need to change something here. You defined it in `Fabulous.XamarinForms.nuspec` and then you depend on it in the extensions.

TimLariviere:
>Will update them manually this time but I'll see what is possible. Maybe just some Regex-magic.

I'm onto something, will make a PR later

azure-pipelines-full.yml
SergejDK: Thought with the default-pipeline it can work but because I wanted to set the Versions in the template it won't. I'll try it with a shell-script - should work.

SergejDK:
@TimLariviere 
Or should I maybe make the script working with params -> so each one will be shown as a own step in azure? I don't think that it is needed because they don't do a lot of things. But for better control on azure it could be a good thing....

azure-pipelines-full.yml
```diff
@@ -30,6 +30,7 @@ jobs:
   pool:
     vmImage: 'macOS-10.14'
   steps:
+  - template: default-osx-template.yml
```
TimLariviere: Did you meant "- script: ./azure-pipeline-osx-setup.sh"?

SergejDK: oh forgot to change that....

SergejDK:
Merging it now!

PR #512, opened 2019-07-09
Change Fabulous.XamarinForms package generation to remove need for nuspec files

TimLariviere:
Due to the fact we generate part of the Fabulous.XamarinForms code, we needed to split the project in 2:
- Fabulous.XamarinForms (all utility functions and classes)
- Fabulous.XamarinForms.Controls (the generated View API)

But to avoid referencing too much packages and since it made no sense to have either F.XF or F.XF.Controls alone, I made a nuspec file to embed both dlls in a single nupkg

This lead to an issue where F.XF extensions were depending on "Fabulous.XamarinForms.Controls" which was not an existing NuGet package.
So I had to make a nuspec file for them as well

This way of generating packages causes some issues:
- Potential errors when writing nuspec file (invalid/missing dependency, wrong package version, etc.)
- Hard to update (until now, package versions were defined in one place Packages.targets, now there is 5 more files to update)

So in order to simplify this, I changed the semantic of our projects:
- Fabulous.XamarinForms.Core (previously Fabulous.XamarinForms) will contain all utility functions and classes like today, so the Generator can still use reflection on it to generate the code
- Fabulous.XamarinForms (previously Fabulous.XamarinForms.Controls) will contain links to all the files of Fabulous.XamarinForms.Core + the generated code (no dependency on F.XF.Core)

This allows to use `dotnet pack Fabulous.XamarinForms` that will generate a single package with a single `Fabulous.XamarinForms.dll` that will contain everything.

All nuspec files are gone (except for templates)

Fabulous.XamarinForms.Core will only be used by the generator itself.
All other projects should depend on Fabulous.XamarinForms


TimLariviere:
@SergejDK I did have a few issues like that, but it was solved by restarting Visual Studio.
Seems it doesn't like switching version of XF without restarting.

Could you try to restart Visual Studio to see if the error go away?

SergejDK:
Yes went away... Thought I restarted... 

Although one error is remaining in all Files:
`type Fabulous.XamarinForms.View with` - XamarinForms not defined

TimLariviere:
Weird. I don't have the issue, nor Azure Pipelines...
Is it just IntelliSense telling you that, or it fails to build too?

SergejDK:
Builds are working.
Can run the apps.
Seems like Intellisense/VS does not want to work :/

Thought it was caused by the changes but it is more likely a VS problem! Maybe the update will help...

So good to go with this pr

PR #517, opened 2019-07-11
Add ProduceReferenceAssembly to .NET Standard projects

TimLariviere:
Seen in the Xamarin Developer Summit live stream:
Screenshot 2019-07-11 at 5 32 08 PM

This will help boost a little the compilation time when working on the Fabulous solution


SergejDK:
Great stuff. Wasn't able to watch the stream... But yes this could help! 

Couldn't test the changes yet but it looks good to me

SergejDK:
@TimLariviere 
did they say which issues can occure with iOS?

TimLariviere:
No he didn't say. Just that it will be fixed very soon.
I guess the issues might be build errors when we will try to compile the iOS projects.

That's why I didn't merged it yet. We might want to wait a little (and play with it too!).

SergejDK:
> That's why I didn't merged it yet. We might want to wait a little (and play with it too!).

Definitively not merging until it works without the main issues. Would be interesting to see the difference on Fabimals and AllControls. I think there we can see a bit.

Issue #520, opened 2019-07-15
Feature request: Support for Context Actions in ListView/TableView


katatunix:
I would like to have `Context Actions` for each item/row in a `ListView` or `TableView`. From the [document](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/listview/interactivity#Context_Actions) of Xamarin.Forms `ListView`, it supports `Context Actions`. So it would be great if we can have it in Fabulous too.

TimLariviere:
Dev notes: The way I can see it implemented is by creating a fake attached property (e.g. "ViewCellContextActions") in the ViewAPI for all Elements (like Created/Ref)
Then in ViewElementCell, we will be able to read it and correctly set the ContextActions property of the ViewCell

TimLariviere:
`ListView` has been changed to accept cell-derived types instead of directly labels and such to allow accessing the ContextActions.
Here's a documentation about it: https://fsprojects.github.io/Fabulous/Fabulous.XamarinForms/migration-guide-to-0.50.html#listview-and-listviewgrouped-now-require-items-and-group-headers-to-be-cells-adds-support-for-textcell--imagecell--switchcell--entrycell--viewcell

It will be released with v0.50 which will arrive in the coming days

Issue #519, opened 2019-07-12
View.Checkbox calls `checkedChanged` when ListView updates


Xatter:
I am trying to create a Todo list style example app with Fabulous using the new Checkbox control in 0.41.0. 

```
open System
open System.Diagnostics

open Fabulous
open Fabulous.XamarinForms
open Xamarin.Forms

module App = 
    type Todo = { 
        id : Guid
        title : string
        completed: bool
    }
    with static member Zero = {
            id = Guid.Empty
            title = String.Empty
            completed = false
        }

    type Model = {
        newItemTitle : string
        todos: Todo list
    }
    with static member Zero = {
            newItemTitle = String.Empty
            todos = [
                {id=Guid.NewGuid(); title="Item3"; completed=false}
                {id=Guid.NewGuid(); title="Item2"; completed=false}
                {id=Guid.NewGuid(); title="Item1"; completed=false}
            ]
        }

    type Msg = 
        | Add of Todo
        | NewItemTitle of string
        | ClearTitle
        | ToggleCompleted of Guid * Boolean

    let init () = Model.Zero, Cmd.none

    let update msg model =
        match msg with
        | Add todo -> { model with todos=todo::model.todos }, (Cmd.ofMsg ClearTitle)
        | NewItemTitle title -> { model with newItemTitle=title }, Cmd.none
        | ClearTitle -> { model with newItemTitle=String.Empty }, Cmd.none
        | ToggleCompleted (tid, value) -> 
            let newTodos =
                model.todos
                |> List.map(fun t ->
                    if t.id <> tid then t
                    else
                        { t with completed = value }
                )
            { model with todos=newTodos }, Cmd.none

    let view (model: Model) dispatch =
        View.ContentPage(useSafeArea=true,
            content = View.StackLayout(
                children=[
                    View.StackLayout(
                        orientation=StackOrientation.Horizontal,
                        children=[
                            dependsOn model.newItemTitle (fun model title -> View.Entry(
                                    text=title,
                                    horizontalOptions = LayoutOptions.FillAndExpand,
                                    textChanged = debounce 250 (fun e -> dispatch (NewItemTitle e.NewTextValue))
                            ))
                            View.Button(
                                text = "+",
                                widthRequest=50.0,
                                horizontalOptions=LayoutOptions.End,
                                command = (fun () -> dispatch (Add { Todo.Zero with id=Guid.NewGuid();title=model.newItemTitle }))
                            )
                        ]
                    )

[...]

Issue #421, opened 2019-05-24
[Generator] Add support for the ObsoleteAttribute


TimLariviere:
Xamarin.Forms 4.0 marked a few properties as obsolete.
To avoid making breaking changes, we should still continue to provide support for these properties, but we need a way to mark them as obsolete in Fabulous as well.

This, for 2 reasons:
- Report to the user that XF obsoleted the property they're using (and maybe later XF and Fabulous will remove this property)
- Avoid having warnings for these obsolete properties when compiling Fabulous

>Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(2985,37): warning FS0044: This construct is deprecated. Image is obsolete as of 4.0.0. Please use ImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(2986,37): warning FS0044: This construct is deprecated. Image is obsolete as of 4.0.0. Please use ImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(9686,37): warning FS0044: This construct is deprecated. BackgroundImage is obsolete as of 4.0.0. Please use BackgroundImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(9687,37): warning FS0044: This construct is deprecated. BackgroundImage is obsolete as of 4.0.0. Please use BackgroundImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(9691,37): warning FS0044: This construct is deprecated. Icon is obsolete as of 4.0.0. Please use IconImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(9692,37): warning FS0044: This construct is deprecated. Icon is obsolete as of 4.0.0. Please use IconImageSource instead.
Fsc: /Git/Github/Fabulous/src/Fabulous.Core/Xamarin.Forms.Core.fs(10222,45): warning FS0044: This construct is deprecated. SetTitleIcon is obsolete as of 4.0.0. Please use SetTitleIconImageSource instead.
[...]

PR #530, opened 2019-07-29
[WIP] Generator v2 and changes to View API

TimLariviere:
Complete rewrite of the Generator.

Will take care of #421, #304, #223, #519, #520, #554, #570 

The initial idea was to rewrite it to add a distinction between members (properties, events, attached properties) because events and attached properties had some issues (see issues above)

But I also wanted to simplify the bindings file because there was a lot of repeated code both in the json and the ViewConverters.fs. Guessing default values was hard too.
So I wanted to automate this.

And lastly, with the clear distinction between Fabulous and Fabulous.XamarinForms, it was the occasion to make the Generator work with everything and not just Xamarin.Forms.

It now has 3 parts in it:
- **AssemblyReader**: Uses both Mono.Cecil and System.Reflection to figure out the maximum of controls, properties, events and attached properties available in one or more assemblies.
- **Binder**: Uses the result of the AssemblyReader and the file provided by the developer (e.g. our Xamarin.Forms.Core.json) to produce the final bindings result (all authorized types, properties, events, etc. in the required order and the required overrides). The Binder will also try to optimize the bindings to reduce the generated code and allocations at runtime.
- **CodeGenerator**: The same existing part as before. Will use the final bindings to generate the corresponding code.

Also Fabulous.Generator (.NET Core app) became Fabulous.CodeGen (.NET Standard 2.0 lib).
A new projet Fabulous.XamarinForms.Generator has been created to make use of Fabulous.CodeGen.

This change allows to make Fabulous.CodeGen completely independent of Xamarin.Forms.
Also this shows how to write our own generator for a given framework (e.g. Fabulous.WPF) and potentially extend it or change the way the code is generated.

Old JSON format: https://github.com/fsprojects/Fabulous/blob/master/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
New JSON format: https://github.com/fsprojects/Fabulous/blob/57c5ad52d6b4553f984cb51046cb3a99c3dc6a06/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.v2.json

What remains:

- [x] Ensure PrimaryConstructorMember is correctly set for all types
- [x] Ensure CanBeInstantiated is correctly set for all types
- [x] Move the call to the update method of the inherited control after the unsubscribing of event, otherwise inherited properties related to events in super type will trigger unwanted events
- [x] Debug samples (CollectionView doesn't work for the moment)
- [x] Merge current develop into this branch
- [x] Simplify `InputTypes`
- [x] Release a preview build to let people play with it
- [x] Replace PrimaryConstructorMember with PrimaryConstructorMembers to allow defining multiple members, then every other members should be sorted alphabetically
- [x] Find a way to replace `Resolver.getAllPropertiesWithBindingFieldForType` because it is highly specific to XAML-based frameworks
- [x] Move Fabulous.CodeGen into his own folder (like F.XF, F.StaticView)
- [x] Clean up ViewConverters and ViewUpdaters (lots of dead or duplicated code)
- [x] Write migration guide from v0.43


TimLariviere:
@SergejDK @dsyme I can't think of a good way to handle attached properties like `Shell.BackgroundColor`.

Currently, the way it's implemented in Fabulous is through the `attached` property on collection members (like `items`, `children`) in the bindings file.
https://github.com/fsprojects/Fabulous/blob/98c44abe67b0d3bf61390a408ad94d5139b6ca0f/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json#L758-L790

This results in updating those attached properties every time the collection is updated (through `updateCollectionGeneric`).

It works perfectly, but only if the attached property only applies to the direct children of a declaring control (like Grid or AbsoluteLayout for instance).

But some of these attached properties can apply to different levels within the hierarchy of the declaring control.
For instance, a `ContentPage` can have a `Shell.SearchHandler` attached property applied to it but not be the direct child of `Shell`.
Instead it will be like `Shell > Tab > ShellContent > ContentPage`.

Today, I solved this by creating fake properties on the target type, instead of the declaring type.
But if we want to let library authors to use the Generator to wrap their controls libraries automatically, this won't work because they won't be able to add those fake properties to base target types (like Xamarin.Forms.Element) since we've already generated those.

TimLariviere:
Maybe we can stay on the current `attached` property way of doing.

After discussing it with some XAML developers, it appears other cases are negligible enough to not include them in the Generator. (people will need to use `created` instead) 

In the case of `Shell`, I think it can be sorted out by using a mix of `attached` and fake properties.
Shell has a limited sets of authorized child controls, so we can add this `Shell.SearchHandler` (and others) as a fake property on these different controls directly.

vshapenko:
> Maybe we can stay on the current `attached` property way of doing.
> 
> After discussing it with some XAML developers, it appears other cases are negligible enough to not include them in the Generator. (people will need to use `created` instead)
> 
> In the case of `Shell`, I think it can be sorted out by using a mix of `attached` and fake properties.
> Shell has a limited sets of authorized child controls, so we can add this `Shell.SearchHandler` (and others) as a fake property on these different controls directly.

@TimLariviere  In case of attached properties that change the parent element of visual tree, we could use a handlers of such type: unit->Color, for example. Also we can use an events as a transport for parent elements getting notified about their properties changed. So , for example , i may have an Event<obj> list of events which are passed down to child elements. 
Another way, i think, would be passing a collection of string*ViewElement  properties to child view. This would help to keep up a property precedence.

TimLariviere:
A little status update:
Most of the work has been done, and everything compiles! \o/

- [ ] Ensure PrimaryConstructorMember is correctly set for all types
- [ ] Ensure CanBeInstantiated is correctly set for all types
- [ ] Debug samples (CollectionView doesn't work for the moment)
- [ ] Merge current develop into this branch
- [ ] Move Fabulous.CodeGen into his own folder (like F.XF, F.StaticView)
- [ ] Write unit tests for Fabulous.CodeGen
- [ ] Write unit tests for Fabulous.XamarinForms.Generator
- [ ] Write technical docs for Fabulous.CodeGen
- [ ] Write user docs for Fabulous.CodeGen
- [ ] Write user docs for Fabulous.XamarinForms.Generator
- [ ] Release a preview build to let people play with it
- [ ] Clean up ViewConverters and ViewUpdaters (lots of dead or duplicated code)

TimLariviere:
>Do you try to achieve a better type-safety with alle the 'Fontsize.Value' etc?

@SergejDK Yes.
As discussed in #139, Don recommended to wait for F# 5.0 with its possible ` type-tagged union types (string|ImageSource) (erased to obj)`.

But since it doesn't seem to be around the corner (I guess it will be available along with .NET 5, in November 2020), I've tried to use discriminated unions to replace the loosely-typed `obj` properties.

That said, I'm not satisfied with the result.
There's actually a lot of name collisions forcing usage of the union name, which make it less practical.

So it might not make it out in this PR, depends on whether we can find a better way for it.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -1,2837 +1,2408 @@
...
-          "modelType": "(int * int) option",
-          "convToValue": "(function None -> null | Some (i,j) -> let items = target.ItemsSource :?> System.Collections.Generic.IList<ListGroupData> in (if i >= 0 && i < items.Count then (let items2 = items.[i] in if j >= 0 && j < items2.Count then items2.[j] else null) else null))"
-        },
-        {
-          "name": "SeparatorVisibility",
-          "defaultValue": "Xamarin.Forms.SeparatorVisibility.Default"
-        },
-        {
-          "name": "SeparatorColor",
-          "defaultValue": "Xamarin.Forms.Color.D
```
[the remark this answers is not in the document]
TimLariviere: It's related to the ordering of the members when writing `View.Entry(...)`.

Today, the order of the constructor members is dependent of the order of declaration in the JSON file.

For example, we have declared `Text` first for `Label`.
This allows us to write `View.Label "Hello World"` without specifying `text=`. 

But this JSON-dependent order has several limitations:
- Harder to maintain the JSON because the properties are not declared alphabetically, whereas MSDN shows them alphabetically sorted.
- Not possible to put before a member of an inherited type if the new type also has its own properties. `View.SomeSuperType(propSuperA, propSuperB, propInheritedA, propInheritedB)`

So I made the order of declaration not relevant, thus we can make it alphabetically sorted.

From what I could see among all the Fabulous apps, only the first constructor member seemed to be used implicitly.
So I created added `PrimaryConstructorMember` field. It allows us to declare which property/event (inherited or not) will be the first member when using the control.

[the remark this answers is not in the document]
SergejDK: Great stuff! 
I didn't think about this.

SergejDK:
> There's actually a lot of name collisions forcing usage of the union name, which make it less practical.

You are right and it is harder to write. Will try to think about a good solution on this

TimLariviere:
@SergejDK I tried to reduce the new `InputTypes` (AllSides and such) to a minimum for now.
To get rid of the `obj` typing in much of the cases, I did the following changes:

- Removed the `InputTypes.Thickness` DU and instead directly use the struct `Xamarin.Forms.Thickness` which does essentially the same thing.

E.g.

```fsharp
// All sides at 20.
View.Label(padding = Thickness 20.)

// Left/Right at 15.; Top/Bottom at 20.
View.Label(padding = Thickness (15., 20.))

// Left at 10.; Top at 15.; Right at 20.; Bottom at 25.
View.Label(padding = Thickness (10., 15., 20., 25.))
```

We still need to add the `Thickness` struct name compared to now (`View.Label(padding = 20.)`) to get rid of the `obj` type.

- Removed `InputTypes.StyleClass` and instead directly use `string list`

```fsharp
// One class
View.Label(styleClass = [ "red-label" ])

// More classes
View.Label(styleClass = [ "red-label"; "bold-label"; "title-label" ])
```

- Renamed `InputTypes.RowOrColumn` to `InputTypes.Dimension`; Cases name should be unique enough to allow use without DU name

```fsharp
View.Grid(
    rowdefs = [ Auto; Star; Stars 0.5; Absolute 50. ],
    coldefs = [ Auto; Star; Stars 0.5; Absolute 50. ]
)
```

- Renamed `InputTypes.FontSize.Value` to `InputTypes.FontSize.FontSize` to avoid name collision and allow use without DU name

```fsharp
// Define a font with a numeric size
View.Label(fontSize = FontSize 15.)

// Define a font with a named size
View.Label(fontSize = Named NamedSize.Large)
```

- Removed `InputTypes.Accelerator` and instead directly use `string` (`Xamarin.Forms.Accelerator` is not supposed to be instantiated any other way)

```fsharp
View.MenuItem(accelerator = "ctrl+A")
```

- Removed `InputTypes.ViewOrText` and instead only require a `ViewElement`

SergejDK:
> We still need to add the Thickness struct name compared to now (View.Label(padding = 20.)) to get rid of the obj type.

I think this is okay because XF handles it the same way?
We reference the XF Doc a lot and for the guys coming from XF this is a good solution.

The rest seems good. :)

Still need to pull this PR and test it for myself

SergejDK:
@TimLariviere 
Is the problem of #465 currently possible in alpha?

TimLariviere:
With the breaking changes to the DSL of this PR, I've also added another one. ListView and Cells.
Instead of implicitly create a ViewCell for each item in a ListView, I've made them explicit.

*Old:*
```fsharp
View.ListView(items = [
    for i = 0 to 1000 do
        yield View.Label(text = sprintf "Test %i" i)
])
```

*New:*
```fsharp
View.ListView(items = [
    for i = 0 to 1000 do
        yield View.ViewCell(
            content = View.Label(text = sprintf "Test %i" i)
        )
])
```

While this may seem to just add more code on the user's end, this will add a bunch of new things:
- Closer to Xamarin.Forms (so docs and all)
- Possibility to use the efficient TextCell, ImageCell, SwitchCell, EntryCell instead of just ViewCell
- Possibility to access those cells specific properties (like ContextActions #520)

*Mix-match cells:*
```fsharp
View.ListView(items = [
    View.TextCell(text = "Hello", details = "World")
    View.SwitchCell(text = "Enable", on = model.On)
    View.EntryCell(text = model.EntryText, textChanged = (fun e -> dispatch EntryChanged e.NewText))
    View.ViewCell(content = View.StackLayout(items = [
        View.Label("Custom Cell")
    ])
])
```

*Context actions:*
```fsharp
View.ListView(items = [
    for i = 0 to 1000 do
        yield View.TextCell(
            text = sprintf "Test %i" i,
            contextActions = [
                View.MenuItem(text = "Edit")
                View.MenuItem(text = "Delete")
            ]
        )
])
```

TimLariviere:
For info, I successfully used the new CodeGen lib to create Fabulous.WPF and port the CounterApp sample. :)
https://github.com/TimLariviere/Fabulous.WPF

.gitignore
```diff
@@ -20,3 +20,21 @@ Fabulous\.XamarinForms/samples/TicTacToe/TicTacToe/TicTacToe\.args\.txt
...
+
+bindings\.json
+
+expanded-bindings\.json
+
+Fabulous\.XamarinForms/src/Fabulous\.XamarinForms/Xamarin\.Forms\.Core\.v2\.fs
+
+optimized-bindings\.json
+
+reader-data\.json
```
TimLariviere: Clean gitignore file

Fabulous.CodeGen/src/Fabulous.CodeGen/AssemblyReader/AssemblyResolver.fs
```diff
@@ -1,8 +1,8 @@
-// Copyright 2018 Fabulous contributors. See LICENSE.md for license.
-namespace Fabulous.Generator
+// Copyright 2018-2019 Fabulous contributors. See LICENSE.md for license.
```
TimLariviere: Make sure all files have the copyright header

TimLariviere:
For those interested, I wrote a migration guide: https://github.com/fsprojects/Fabulous/blob/fd78a9e82f03b3c230528e86dd145b7526f90931/docs/Fabulous.XamarinForms/migration-guide-to-0.50.md

Fabulous.XamarinForms/samples/AllControls/AllControls/AllControls.fs
```diff
@@ -597,11 +598,11 @@ module App =
                dependsOn () (fun model () -> 
                  View.NonScrollingContentPage("Image", 
                      [ View.Label(text="Image (URL):")
-                       View.Image(source="http://upload.wikimedia.org/wikipedia/commons/thumb/f/fc/Papio_anubis_%28Serengeti%2C_2009%29.jpg/200px-Papio_anubis_%28Serengeti%2C_2009%29.jpg", 
+                       View.Image(source=Path "http://upload.wikimedia.org/wikipedia/commons/thumb/f/fc/Papio_anubis_%28Serengeti%2C_2009%29.jpg/200px-Papio_anubis_%28Serengeti%2C_2009%29.jpg", 
                            horizontalOptions=LayoutOptions.FillAndExpand,
                            verticalOptions=LayoutOptions.FillAndExpand)
-                       View.Label(text="Image (Embedded):", margin=Thickness(0., 20., 0., 0.))
-                       View.Image(source=ImageSource.FromResource("AllControls.Baboon_Serengeti.jpg", typeof<RootPageKind>.Assembly), 
+                       View.Label(text="Image (Embedded):", margin = Thickness (0., 20., 0., 0.))
```
dsyme: Lovely work on these type safety changes

Fabulous.XamarinForms/samples/AllControls/AllControls/AllControls.fs
```diff
@@ -736,37 +752,37 @@ module App =
...
+                                          .AbsoluteLayoutLayoutBounds(Rectangle(0.0, 0.0, AbsoluteLayout.AutoSize, AbsoluteLayout.AutoSize))
                                       View.Label(text = "Centered", textColor = Color.Black)
-                                          .LayoutFlags(AbsoluteLayoutFlags.PositionProportional)
-                                          .LayoutBounds(Rectangle(0.5, 0.5, AbsoluteLayout.AutoSize, AbsoluteLayout.AutoSize))
+                                          .AbsoluteLayoutLayoutFlags(AbsoluteLayoutFlags.PositionProportional)
+                                          .AbsoluteLayoutLayoutBounds(Rectangle(0.5, 0.5, AbsoluteLayout.AutoSize, AbsoluteLayout.AutoSize))
                                       View.Label(text = "Bottom Right", textColor = Color.Black)
-                                          .LayoutFlags(AbsoluteLayoutFlags.PositionProportional)
-                                          .LayoutBounds(Rectangle(1.0, 1.0, AbsoluteLayout.AutoSize, AbsoluteLayout.AutoSize)) ])
+                                          .AbsoluteLayoutLayoutFlags(AbsoluteLayoutFlags.PositionProportional)
```
dsyme: Shame about the long names, I guess they are unavoidable?

Fabulous.XamarinForms/samples/Calculator/Calculator/Calculator.fs
```diff
@@ -94,9 +94,9 @@ module App =
                 .TextColor(Color.Black) 
 
         View.ContentPage(
-            View.Grid(rowdefs=[ "*"; "*"; "*"; "*"; "*"; "*" ], coldefs=[ "*"; "*"; "*"; "*" ],
+            View.Grid(rowdefs=[ Star; Star; Star; Star; Star; Star ], coldefs=[ Star; Star; Star; Star ],
```
dsyme: Any chance of a more semantically meaningful name than `Star`, perhaps `Fit` or `Proportional` or `Weighted` ...? Just wondering, `Star` might be best I guess

Fabulous.XamarinForms/samples/AllControls/AllControls/AllControls.fs
TimLariviere: I changed the way AttributeKeys are de-duplicated (with their corresponding methods like this one). Names are simplified only if a property (same name, same signature) is used on 2+ controls.

But after updating the samples, I realize it isn't a good idea since it makes those methods name longer for nothing (and breaks existing code).

So I will revert to the previous behavior instead. (making names shorter)

Cases where name collision with different signatures might happen, but hopefully they'll be rare. (in this case, it will be using unique names)

Fabulous.XamarinForms/samples/Calculator/Calculator/Calculator.fs
TimLariviere: While being more semantically meaningful might seem interesting, I think the best for us is to use what people will expect to find.

We're not tied to what Xamarin.Forms/XAML does, and Fabulous.XamarinForms is a good occasion to innovate on this front.

But renaming these names, I'm afraid we'll lose people more than any advantage it might give them.

People (beginners and XAML developers) will search/read docs and stackoverflow, where they will find mostly things on Xamarin.Forms and XAML in general.

So using the same names will be more natural for them.

SergejDK:
@TimLariviere 
Was building some stuff for myself with the latest alpha release and it works pretty well for me! Good job! 

Do we need to do something on the extensions like Map?

TimLariviere:
You mean, to use code gen for extensions?
It is possible to write the bindings file for the extensions, the code will be generated.

The only thing I haven't really found a good way to handle is the inherited properties that come from another bindings file. (like Xamarin.Forms.Core.json)

I guess the easiest way would be to add at the root of the json a "default" inherited type with all the properties. (here we just need a simplified definition of the type since we only need to call its Build and Update methods)

```json
{
    "assemblies": [
        "Xamarin.Forms.Maps.dll"
     ],
    "outputNamespace": "Fabulous.XamarinForms.Maps",
    "baseType": {
        "name": "View",
        "members": [
            { "name": "horizontalOptions", "inputType": "Xamarin.Forms.LayoutOptions" },
            { "name": "verticalOptions", "inputType": "Xamarin.Forms.LayoutOptions" },
            (...)
        ]
    },
    "types": [
        {
            "type": "Xamarin.Forms.Map",
            "properties": [ (...) ]
         }
    ]
}
```

SergejDK:
> Or maybe by adding a reference to the other json file?

If this is possible somehow I would use the reference instead - makes it a lot easier or not?

TimLariviere:
>If this is possible somehow I would use the reference instead - makes it a lot easier or not?

Still needs to check the feasibility. It should be doable.

TimLariviere:
Since the majority of the work has been done (CodeGen + View API changes) and the alpha release proved successful, I'll merge this PR and open a few tickets for tracking the remaining work.

This will let us include support for Xamarin.Forms 4.3 without conflicting.

SergejDK:
> Since the majority of the work has been done (CodeGen + View API changes) and the alpha release proved successful, I'll merge this PR and open a few tickets for tracking the remaining work.
This will let us include support for Xamarin.Forms 4.3 without conflicting.

I think it is okay to merge. I currently don't see any problems in the codebase as in working with the Alpha version. Also you included how to upgrade from 0.4 to 0.5 :)

dsyme:
@TimLariviere Getting this merged is a huge step towards Fabulous 1.0 - great job!

PR #531, opened 2019-07-31
Fixed updateCollectionGeneric

vshapenko:
Closes #518

Issue #535, opened 2019-08-01
Null Reference Exception thrown when updating views


TimLariviere:
Original issue: #518 

While updating the UI when switching between 2 different contents (Label -> ListView), the app throws a NullReferenceException (coming from Xamarin.Forms) but only on some Android devices.

- Affected platforms: Android
- Affected build mode: Debug, Release
- Affected devices:
    - Xiaomi Mi 8
    - Samsung S8
- Unaffected devices:
    - Emulators
    - Xiaomi Redmi Note 7
    - LG G6
    - Google Nexus 5
    - Sony Xperia
    - Motorola G6
    - Xiaomi Redmi 4X
    - Nokia 8

# Stack trace

<details>
<summary>Stack trace for Repro 1</summary>

```
Unable to update view:: System.NullReferenceException: Object reference not set to an instance of an object.
at Xamarin.Forms.Platform.Android.PageRenderer.Xamarin.Forms.Platform.Android.IOrderedTraversalController.UpdateTraversalOrder () [0x000a4] in D:\a\1\s\Xamarin.Forms.Platform.Android\Renderers\PageRenderer.cs:177
at Xamarin.Forms.Platform.Android.VisualElementRenderer1[TElement].UpdateParentPageTraversalOrder () [0x00025] in D:\a\1\s\Xamarin.Forms.Platform.Android\VisualElementRenderer.cs:404 at Xamarin.Forms.Platform.Android.VisualElementRenderer1[TElement].OnElementPropertyChanged (System.Object sender, System.ComponentModel.PropertyChangedEventArgs e) [0x00116] in D:\a\1\s\Xamarin.Forms.Platform.Android\VisualElementRenderer.cs:369
at (wrapper delegate-invoke) .invoke_void_object_PropertyChangedEventArgs(object,System.ComponentModel.PropertyChangedEventArgs)
at Xamarin.Forms.BindableObject.OnPropertyChanged (System.String propertyName) [0x00000] in D:\a\1\s\Xamarin.Forms.Core
BindableObject.cs:211
at Xamarin.Forms.Element.OnPropertyChanged (System.String propertyName) [0x00000] in D:\a\1\s\Xamarin.Forms.Core\Element.cs:353
at Xamarin.Forms.Element.set_Parent (Xamarin.Forms.Element value) [0x00094] in D:\a\1\s\Xamarin.Forms.Core\Element.cs:210
[...]

PR #540, opened 2019-08-03
Temporary fix for #535

Happypig375:
Closes #531


TimLariviere:
Sorry, I know you meant well but we can't accept this pull request either.

I prefer to wait for an official answer from Xamarin.Forms before taking any action since this bug only happens on a fraction of the devices using Android (which itself is only 1 platform among the 7 supported by Fabulous.XamarinForms).

I'm closing the pull request to avoid it being merged by accident.
But we can continue to discuss here.

PR #544, opened 2019-08-09
Make Effects easier

### LINKED ISSUE #457: Allow effects to be specified on View declaration
This would simplify the use of effects in Xamarin.Forms. e.g.
```fs
View.Label(effects = ["SomeResolutionGroup.SomeEffectName"])
```
instead of
```fs
View.Label(created = fun e -> e.Effects.Add <| Effect.Resolve "SomeResolutionGroup.SomeEffectName")
```

SergejDK:
This should make the use of effects a lot easier.
Fix #457 

@TimLariviere 
couldn't test this right now but I think this should work as expected.

### DIFF
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
@@ -833,8 +833,7 @@ module Converters =
 
     let updateEffects (prevCollOpt: ViewElement array voption) (collOpt: ViewElement array voption) (target: Xamarin.Forms.Element) =
         let create (viewElement: ViewElement) =
-            let effect = viewElement.Create()
-            match effect with
+            match viewElement.Create() with
             | :? CustomEffect as customEffect -> Effect.Resolve(customEffect.Name)
-            | _ -> effect :?> Xamarin.Forms.Effect
+            | effect -> effect :?> Xamarin.Forms.Effect
         updateCollectionGeneric prevCollOpt collOpt target.Effects create (fun _ _ _ -> ()) ViewHelpers.canReuseView updateChild
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
@@ -830,4 +830,19 @@ module Converters =
         | ValueNone, ValueNone -> ()
         | _, ValueSome currValue -> setUseSafeArea currValue
         | ValueSome _, ValueNone -> setUseSafeArea false
+
+    let updateEffects (prevValueOpt: string[] voption) (currValueOpt: string[] voption) (target: Xamarin.Forms.Element) =
+        match prevValueOpt, currValueOpt with
+        | ValueSome prevValue, ValueSome currValue when prevValue = currValue -> ()
+        | ValueNone, ValueNone -> ()
+        | _, ValueSome currValue -> 
+            let createEffect effectName = Effect.Resolve effectName
+            currValue 
+            |> Array.iter (fun v -> 
+                let effect = createEffect v
+                match target.Effects.Contains(effect) with
+                | true -> ()
+                | false -> target.Effects.Add(effect)
+            )
+        | ValueSome _, ValueNone -> target.Effects.Clear()
         
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
@@ -832,7 +832,9 @@ module Converters =
         | ValueSome _, ValueNone -> setUseSafeArea false
 
     let updateEffects (prevCollOpt: ViewElement array voption) (collOpt: ViewElement array voption) (target: Xamarin.Forms.Element) =
-        let create (viewElement: ViewElement) = 
-            let customEffect = viewElement.Create() :?> CustomEffect
-            Effect.Resolve(customEffect.Name)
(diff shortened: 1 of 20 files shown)


Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
```diff
@@ -830,4 +830,19 @@ module Converters =
         | ValueNone, ValueNone -> ()
         | _, ValueSome currValue -> setUseSafeArea currValue
         | ValueSome _, ValueNone -> setUseSafeArea false
+
+    let updateEffects (prevValueOpt: string[] voption) (currValueOpt: string[] voption) (target: Xamarin.Forms.Element) =
```
TimLariviere: I think you can call `updateCollectionGeneric` directly in this function, instead of writing this pattern matching.
Currently this function doesn't clean up removed effects when the new value is not null.

Example:
`prev = View.Element(effects=["A"; "B"])`
`curr = View.Element(effects=["A"])`

SergejDK: yes. forgot about it and currently trying to make it work with updatecollectiongeneric. 

Have a little problem on my side that "string is not compatible with Viewelement".

I thjink it is because my createfunction returns an Effect and not a ViewElement.Effect

Do you think we should just make a ViewElement.effect? 

Changed it to sth. like this:

```fsharp
    let updateEffects (prevValueOpt: string array voption) (currValueOpt: string array voption) (target: Xamarin.Forms.Element) =
        let createEffect effectName = Effect.Resolve effectName
        updateCollectionGeneric prevValueOpt currValueOpt target.Effects createEffect (fun _ _ _ -> ()) ViewHelpers.canReuseView updateChild
```

Happypig375:
I just discovered that effects can have custom properties through inheriting from `RoutingEffect` and finding that effect from the corresponding platform effect. Maybe it is better to keep the type as `Effect list` after all.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewConverters.fs
TimLariviere: >Do you think we should just make a ViewElement.effect?

Yes, seems like we should.
This will be necessary for @Happypig375's comment where effects can have parameters.

We can have a default `View.Effect(name=XXX)` that will represent an effect with no parameter.
Then a classic ViewElement creation will be used for effects with parameters.

SergejDK:
We have to create a CustomControl to include a new member and work with `updateCollectionGeneric` which would look like this:

```fsharp
type CustomEffect() =
    inherit Effect() // this will be an error because the constructor is internal
    
    member val Name = "" with get, set
```

So because this gives an error we could go with:
```fsharp
type CustomEffect() =
    inherit RoutingEffect("") 
    
    member val Name = "" with get, set
```

BUT normally it does not matter we can just say we make a type with the Name because the resolve method is static.

So in conclusion:
```fsharp
type CustomEffect() =
    member val Name = "" with get, set
```

The updateFunction for the Effects would be:
```fsharp
    let updateEffects (prevCollOpt: ViewElement array voption) (collOpt: ViewElement array voption) (target: Xamarin.Forms.Element) =
        let create (viewElement: ViewElement) = 
            let customEffect = viewElement.Create() :?> CustomEffect
            Effect.Resolve(customEffect.Name)
        updateCollectionGeneric prevCollOpt collOpt target.Effects create (fun _ _ _ -> ()) ViewHelpers.canReuseView updateChild
```

To use this in the app the call is the following:
```fsharp
View.Button(effects = [View.Effect("EffectName")])
```

I don't like the part with the CustomEffect with only one member.... 

@TimLariviere  
Do you think we should go with this solution or something better?

TimLariviere:
@SergejDK Maybe we should simply provide a wrapper for `RoutingEffect` in the `Xamarin.Forms.Core.json` file?

It seems to do exactly what we're trying to achieve with `CustomEffect` and is also inheriting from `Xamarin.Forms.Effect` which will greatly simplify our `updateEffects` (maybe even remove it entirely?).

The only thing will be to find a way to pass this `Name`/`EffectId` field as a constructor parameter.
Also this would require to disable reuse when the `Name`/`EffectId` changes.

SergejDK:
@TimLariviere 
currently trying to understand how to make a custom constructor with the generator.
Found the `ConstructorType` etc. but I need to figure out how to put it in the json correctly.

Any suggestions? You are rewriting it so you clearly have more knowledge about it :)

TimLariviere:
@SergejDK Well while refactoring the Generator, I noticed that even though it has some code for custom constructors, it isn't fully implemented.
And implementing it would certainly have a big impact on Fabulous (ViewElement will need to be modified)

https://github.com/fsprojects/Fabulous/blob/79c5df748fff7a108dfbcbf8609cb2265a8fddc7/tools/Generator/CodeGenerator.fs#L81-L102

This `HasCustomConstructor` means that no public parameterless constructor has been found on the type (via reflection). So `RoutingEffect` will always throw an exception in this case... (can't create RoutingEffect)

We will need to find a solution to that problem, but I prefer to do that only once #530 is merged.

In the meantime, we can go with your `CustomEffect + updateEffects` solution

SergejDK:
Then I was correct that there is sth. not completly done but great that you saw this!

We definetly should make constructors available after you are ready with the generator :)

SergejDK:
@TimLariviere 
wanted to include it in the docs, too. 
Is the `views-styling.md` the correct one for this?

TimLariviere:
Hmm, effects are not necessarily related to styling.

I think it would be better to create a new page just for Effects.

This way, we will be able to add more details on how to use Effects (how to use custom effects with custom properties?) as well as add some details on potential built-in XF effects.

PR #557, opened 2019-09-10
[WIP] Updated Android template project file

### LINKED ISSUE #542: Check dependencies for the Android template
The Android template seems to be missing some dlls required by Xamarin.Forms. (`Support something something LiveData.dll` and `Java.Interop`)

When updating the Xamarin.Forms package, they get added to the project even though XF doesn't update its Android dependencies.

We need to check that.

TimLariviere:
Fixes #542


TimLariviere:
Will be done in an update for targeting Android 10

PR #559, opened 2019-09-11
Upgrade for Fabulous.XamarinForms to 4.2

### LINKED ISSUE #459: [Samples][GTK] Allcontrols/TicTacToe
TicTacToe/AllControls GTK throughs a nullreferenceexception at start.

### LINKED ISSUE #556: Upgrade Xamrin.Forms to 4.2.0.709249
We can update XF from 4.1 to 4.2.

It was released 9 days ago so I think we can do the upgrade. [https://www.nuget.org/packages/Xamarin.Forms/](https://www.nuget.org/packages/Xamarin.Forms/)

SergejDK:
This PR 
fix #556 
and fix #459  (need to test but should be working now)

---

Currently getting an error on "GroupableItemsViewViewer"  because it is not defined.

### DIFF
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
@@ -2326,15 +2326,7 @@
           "defaultValue": "null"
         },
         {
-          "name": "FooterTemplate",
-          "defaultValue": "null"
-        },
-        {
           "name": "Header",
-          "defaultValue": "null"
-        },
-        {
-          "name": "HeaderTemplate",
           "defaultValue": "null"
         },
         {
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
@@ -2776,6 +2776,25 @@
       ]
     },
     {
+      "name": "Xamarin.Forms.GroupableItemsView",
+      "members": [
+        {
+          "name": "GroupFooterTemplate",
+          "defaultValue": "null"
+        },
+        {
+          "name": "GroupHeaderTemplate",
+          "defaultValue": "null"
+        },
+        {
+          "name": "IsGrouped",
+          "defaultValue": "false",
+          "inputType": "bool",
+          "modelType": "bool"
+        }
+      ]
+    },
+    {
       "name": "Xamarin.Forms.ShellContent",
       "members": [
         {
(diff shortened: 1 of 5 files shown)


Fabulous.XamarinForms/templates/content/blank/.template.config/template.json
```diff
@@ -156,7 +156,7 @@
       "type": "parameter",
       "dataType": "string",
       "replaces": "XamarinFormsSdk",
-      "defaultValue": "4.1.0.555618"
+      "defaultValue": "4.2.0.778463"
```
TimLariviere: Make sure to target the lowest possible version of Xamarin.Forms 4.2 (4.2.0.709249) ;)

SergejDK:
@TimLariviere 
upgrade is done. Tested it on OSX and Windows - Will create a simple Linux machine to have this setup, too.

The fix for the GTK issue with the Navigationpage is gone, too. I found a new issue but I don't know if this is caused by us or XF itself:

Bildschirmfoto 2019-09-13 um 21 49 23


The `&apos;` seems to be a encoding issue of XF. I know that there was a bug with this. 
Do you know something about this?

SergejDK:
@TimLariviere 
the mentioned issue is caused by xamarin.forms. So we can ignore it for the time being.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2316,6 +2320,59 @@
           "inputType": "obj * obj * Xamarin.Forms.ScrollToPosition * Fabulous.XamarinForms.AnimationKind",
           "modelType": "obj * obj * Xamarin.Forms.ScrollToPosition * Fabulous.XamarinForms.AnimationKind",
           "updateCode": "(fun _ curr target -> triggerScrollTo curr target)"
+        },
+        {
+          "name": "Footer",
+          "defaultValue": "null"
+        },
+        {
+          "name": "FooterTemplate",
```
TimLariviere: Could you declare only the direct property and not the Template one (here `Footer` vs `FooterTemplate`, and the others as well) please ?

We don't provide direct support for Template properties (DataTemplate, etc.).

Instead we only expose one property that accepts a ViewElement that we put directly in the "real" property (e.g. Footer) since we don't need to use BindingContext.

SergejDK: So I only have to delete those parts, correct?

TimLariviere: Yes

PR #149, opened 2018-07-31
Documentation + Sample for DisplayAlert/DisplayActionSheet

TimLariviere:

PR #576, opened 2019-10-29
Add support for Xamarin.Forms 4.3

TimLariviere:
Some of the changes have already been added to F.XF 0.43 to avoid getting runtime errors when targeting XF 4.3.

This PR contains the remaining controls and properties changes.

~I still need to add a sample to AllControls for the new control RefreshView.~ Done

Link: https://docs.microsoft.com/en-us/xamarin/xamarin-forms/release-notes/4.3/4.3.0-api

### DIFF
--- a/Fabulous.CodeGen/src/Fabulous.CodeGen/AssemblyReader/Converters.fs
+++ b/Fabulous.CodeGen/src/Fabulous.CodeGen/AssemblyReader/Converters.fs
@@ -2,6 +2,7 @@
 namespace Fabulous.CodeGen.AssemblyReader
 
 open System
+open System.Globalization
 
 module Converters =
     /// Converts the type name to another type name (e.g. System.Boolean => bool)
@@ -26,8 +27,8 @@ module Converters =
         | "System.Collections.IList" -> "obj list"
         | _ -> typeName
         
-    let inline numberWithDecimalsToString literal v =
-        let str = v.ToString()
+    let inline numberWithDecimalsToString literal (v: 'T when 'T :> IConvertible) =
+        let str = v.ToString(CultureInfo.InvariantCulture)
         let separator = if not (str.Contains(".")) then "." else ""
         str + separator + literal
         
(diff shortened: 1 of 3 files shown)


TimLariviere:
@SergejDK Is it ok for you?

SergejDK:
I'm going to check it this evening.

SergejDK:
@TimLariviere 

tried to build on Windows 10. Got a bunch of the following error on the 'RunFabulousCodeGenTests':
```fsharp
X numberWithDecimalsToString should return string representation of number with decimals and type literal<Double,String>(1.5d,"","1.5") [75ms]
  Fehlermeldung:
     Expected string length 3 but was 4. Strings differ at index 1.
  Expected: "1.5"
  But was:  "1,5."
```

It is always the problem between ',' and '.'

TimLariviere:
Ah good catch! Didn't use an invariant culture it seems. 😄 
Will fix that

TimLariviere:
@SergejDK This should fix the issue you found

TimLariviere:
>IsContextActionsLegacyModeEnabled

I hesitated to add this one. After reading the docs, it appears to be better to use the new mode.
Also we haven't released support for ContextActions for now, so I think we can ignore it.
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/platform/android/viewcell-context-actions

>InternalItemsLayout

This one is marked as internal in the source code, and seems only to be usable by controls inheriting from ItemsView.
For example, `StructuredItemsView` re-exposes it as `ItemsLayout`
https://github.com/xamarin/Xamarin.Forms/blob/master/Xamarin.Forms.Core/Items/StructuredItemsView.cs

>DisplayPromptAsync (Page)

It's a method. So same as #149 

>Command (RefreshView)

This one is open to debate :)
I preferred to expose the event instead of the command, and that for several reasons:
- `Command` is not an explicit name. `Refreshing` (event) is.
- F.XF exposes commands and events the same way, even if they don't work the same underneath (`command = fun () -> dispatch XXX`), so they are identical (except one could set `CanExecute` with a command)
- Command is deeply connected to MVVM, so we might want to avoid that when possible

What do you think?

SergejDK:
This is a difficult part.

Well I agree with you that it is connected to MVVM but still we often need the `command` when a event is not the stuff someone needs.
Personally I would expose this because we can't definetly say it is not needed.  Maybe there is a need to use both properties.

TimLariviere:
The more I think about it, the more I think we should not provide both `refreshing` and `command`.

It is confusing as to when use one or the other (spoilers: there's no real difference in this case).

Also we can't prevent `command` to be triggered when Fabulous changes the `isRefreshing` value underneath, where the event won't react (that was the initial objective of #530)

The only thing where there will be a difference is preventing the refresh to happen.

```fsharp
View.RefreshView(
    command = (fun () -> dispatch Refreshing),
    commandCanExecute = false
)

vs.

View.RefreshView(
    refreshing = (fun () -> dispatch Refreshing),
    isEnabled = false
)
```

The latter for me feels way more explicit.

SergejDK:
In this case I completely agree with you but I think we have to check this on the other viewelements(don't know how many there are) about explicity show command or not.

PR #601, opened 2019-11-11
Cleanup to reduce diff

dsyme:
This cleanup reduces the diff w.r.t. the "adaptive" fork a little by aligning names and a few other small things - it is otherwise innocuous/good.


dsyme:
@TimLariviere I'll merge this directly as it's pretty straight forward

PR #610, opened 2019-11-17
Make apparent that the migration guide to Fabulous.XamarinForms is for v0.40

Happypig375:
To emphasis that it is relatively outdated.


docs/Fabulous.XamarinForms/migration-guide-to-0.40.md
```diff
@@ -3,7 +3,7 @@ Fabulous for Xamarin.Forms - Guide
 
 {% include_relative contents.md %}
 
-Migrating from Fabulous v0.36 to Fabulous.XamarinForms
+Migrating from Fabulous v0.36 to Fabulous v0.40
```
TimLariviere: The key thing is `Fabulous.XamarinForms` here, so I would keep the name:
```suggestion
Migrating from Fabulous v0.36 to Fabulous.XamarinForms v0.40
```

PR #611, opened 2019-11-18
Add proper support for ItemsView.EmptyView

### LINKED ISSUE #608: Setting EmptyView on CollectionView crashes the app on iOS and Android
Title. We caught it in our app, I will try to check if the behavior persists on a small repro on weekends.

TimLariviere:
Closes #608


SergejDK:
@TimLariviere 
do we have an example in `AllControls`to see this?

TimLariviere:
No. But we can easily add one though.

SergejDK:
@TimLariviere 
tested it on multiple simulators and emulators - worked well for me.
Thanks for this :)

PR #614, opened 2019-11-21
Changed Type of itemsLayout for CarouselView

### LINKED ISSUE #612: itemsLayout for View.CarouselView has incorrect type
Currently, `View.CarouselView` has a `itemsLayout` property of `ViewElement` type. It should has type of `LinearItemsLayout` as does `CarouselView` itself.

SergejDK:
Should fix #612 

Tested it on Simulator end Emulator.

### DIFF
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
@@ -189,6 +189,13 @@ type CustomTimePicker() =
         if propertyName = "Time" then
             timeChanged.Trigger(this, this.Time)
 
+/// Itemslayout for CarouselView
+type VerticalLinearItemsLayout() = 
+    inherit LinearItemsLayout(ItemsLayoutOrientation.Vertical)
+
+type HorizontalLinearItemsLayout() = 
+    inherit LinearItemsLayout(ItemsLayoutOrientation.Horizontal)
+
 /////////////////
 /// Pages
 /////////////////
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
@@ -195,7 +195,7 @@
         }
       ],
       "events": [
-        { 
+        {
           "source": "Tapped"
         }
       ]
@@ -1387,7 +1387,7 @@
           "source": "Source"
         }
       ],
-      "events":[
+      "events": [
         {
           "source": "Clicked"
         },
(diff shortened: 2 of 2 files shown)


TimLariviere:
I'm not sure this is a good fix.

`LinearItemsLayout` is a class with several properties, not just a value.
So to correctly support it with view diffing, we need it to be a ViewElement (which it is already by the way)
https://github.com/fsprojects/Fabulous/blob/a0754429261ff15cdc1aae2adfda63ce18e75632/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json#L2612-L2622

If the initial idea was to be able to support static instances like `LinearItemsLayout.Horizontal/Vertical`, then I think we should go with specific ViewElements like we did ListView/ListViewGrouped.

```fsharp
type VerticalLinearItemsLayout() =
    inherit LinearItemsLayout(Orientation.Vertical)
```
```json
{
    "type": "Xamarin.Forms.LinearItemsLayout",
    "canBeInstantiated": false,
    (...)
},
{
    "type": "Fabulous.XamarinForms.VerticalLinearItemsLayout",
    (...)
}
```

We can even do something better and slightly modify the generator to be able to create a custom "Create" function.
So instead of what I wrote before, we would have
```json
{
    "type": "Xamarin.Forms.LinearItemsLayout",
    "name": "VerticalLinearItemsLayout",
    "create": "Xamarin.Forms.LinearItemsLayout.Vertical // initial value instead of instantiating something else"
},
{
    "type": "Xamarin.Forms.LinearItemsLayout",
    "name": "HorizontalLinearItemsLayout",
    "create": "Xamarin.Forms.LinearItemsLayout.Horizontal // initial value instead of instantiating something else"
}
```
It would be doing this
```fsharp
member this.CreateVerticalLinearItemsLayout() =
    Xamarin.Forms.LinearItemsLayout.Vertical // initial value instead of instantiating something else
```
And used like this
```fsharp
View.CarouselView(
    itemsLayout = View.VerticalLinearItemsLayout(itemsSpacing = 5.0)
)
```

TimLariviere:
Actually, I thought about it a little more and we should do the `create` override.
Otherwise we will have side effect.
Since the static fields `LinearItemsLayout.Vertical` and `LinearItemsLayout.Horizontal` are unique instances, doing this `View.VerticalLinearItemsLayout(itemSpacing=5.0)` would apply this change to all controls using VerticalLinearItemsLayout.

We should use our own custom instance.

SergejDK:
I thought the other way around but you are completely right. 
Will try the custom part.

SergejDK:
@TimLariviere 
the call would look sth. like: `itemsLayout=View.VerticalLinearItemsLayout()`

SergejDK:
@TimLariviere 
do you think I need to add something here?

PR #644, opened 2019-12-18
XF Upgrade 4.4

### LINKED ISSUE #641: SwipeView/IndicatorView
In 4.4 Xamarin added support for [SwipeVIew](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/swipeview) and [IndicatorView](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/indicatorview). It would be nice to have them in Fabulous.

SergejDK:
fix #641 

TODO:

- [x] IndicatorView Implementation
- [x] SwipeView Implementation
- [x] SwipeItems Implementation
- [x] Use in AllControls
- [x] CarouselVertical


SergejDK:
@TimLariviere 
Should I add the new Views to AllControls ?

TimLariviere:
>Should I add the new Views to AllControls ?

Yes, please.

TimLariviere:
Seems like there's a new `LinearItemsLayout` value like you added in #614: `CarouselVertical`
https://docs.microsoft.com/en-us/xamarin/xamarin-forms/release-notes/4.4/4.4.0-api#type-changed-xamarinformslinearitemslayout

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -1120,6 +1120,9 @@
         {
           "source": "FlyoutIsPresented"
         },
+        {
+          "source": "FlyoutVerticalScrollMode"
```
TimLariviere: Could you implement `NavBarHasShadow` like we did `NavBarIsVisible`, please?

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -1592,12 +1570,6 @@
           "defaultValue": "true",
           "updateCode": "ViewUpdaters.updateCommand prevSearchCommandOpt currSearchCommandOpt (fun (_target: Xamarin.Forms.SearchBar) -> _target.Text) (fun (target: Xamarin.Forms.SearchBar) cmd -> target.SearchCommand <- cmd)"
         },
-        {
-          "source": "Text"
```
TimLariviere: I guess the corresponding TextChanged will be removed too?

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
       "primaryConstructorMembers": [
         "Items"
       ]
+    },
+    {
+      "type": "Xamarin.Forms.IndicatorView",
+      "name": "Indicator",
```
TimLariviere: Is it wanted you renamed it to `Indicator`?
I think it's safe to just remove the `name` attribute from the JSON.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+          "source": "HideSingle"
+        },
+        {
+          "source": "IndicatorLayout"
+        },
+        {
+          "source": "IndicatorSize"
+        },
+        {
+          "source": "IndicatorTemplate"
```
TimLariviere: Oh, I fear this one will need a custom control to correctly support the DataTemplate

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+        },
+        {
+          "source": "IndicatorShape"
+        },
+        {
+          "source": "ItemsSource"
```
TimLariviere: Same with this one. We will need to come up with a usable way.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+          "source": "Position"
+        },
+        {
+          "source": "SelectedIndicatorColor"
+        }
+      ]
+    },
+    {
+      "type": "Xamarin.Forms.SwipeItem",
+      "name": "SwipeItem",
```
TimLariviere: No need to set the `name`. The generator will pick it up automatically

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+      ],
+      "events": [
+        {
+          "source": "Invoked"
+        }
+      ]
+    },
+    {
+      "type": "Xamarin.Forms.SwipeItemView",
+      "name": "SwipeItemView",
```
TimLariviere: No need to set the name. The generator will pick it up automatically

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+      ],
+      "events": [
+        {
+          "source": "Invoked"
+        }
+      ]
+    },
+    {
+      "type": "Xamarin.Forms.SwipeItems",
+      "name": "SwipeItems",
```
TimLariviere: No need to set the name. The generator will pick it up automatically

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+        }
+      ],
+      "events": [
+        {
+          "source": "Invoked"
+        }
+      ]
+    },
+    {
+      "type": "Xamarin.Forms.SwipeItems",
```
TimLariviere: This one can be a problem for incremental updates. It's a list in itself.
We will need to make sure it behaves correctly.

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+          "source": "CollectionChanged",
+          "inputType": "obj -> unit",
+          "modelType": "System.Collections.Specialized.NotifyCollectionChangedEventHandler",
+          "convertInputToModel": "ViewConverters.makeSwipeItemsChangedEventHandler"
+        }
+      ]
+    },
+    {
+      "type": "Xamarin.Forms.SwipeView",
+      "name": "SwipeView",
```
TimLariviere: No need to set the name. The generator will pick it up automatically

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -2822,6 +2791,137 @@
...
+          "source": "LeftItems"
+        },
+        {
+          "source": "RightItems"
+        },
+        {
+          "source": "TopItems"
+        }
+      ],
+      "events": [
```
TimLariviere: For all those 4 events, I think you can just set `"source": "EventName"`. The generator should be able to figure out the rest. (thanks to `System.EventHandler`/`System.EventHandler<T>`)
(So no need for ViewConverters too)

SergejDK:
@TimLariviere 
Good points :) 

I see that we have an `CustomEntryCell` where the Text property and the textchanged event is used. Should we remove those parts? Is there even a need for a customEntryCell anylonger?

TimLariviere:
>I see that we have an CustomEntryCell where the Text property and the textchanged event is used. Should we remove those parts? Is there even a need for a customEntryCell anylonger?

Yes, we need to keep it.
`EntryCell` doesn't inherit from `InputView` so it declares its own properties and events.
We had to write a `CustomEntryCell` because `EntryCell` only provide an overridable `OnTextChanged` instead of an event that we need.

SergejDK:
@TimLariviere 
I was taking a look on how to use IndicatorView and somehow I think we do not need to use ItemsSource and IndicatorTemplate.

https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/indicatorview

Normally the method `ItemsSourceBy` is used with a reference to a CarouselView. 

What do you think about exposing this function as a event and accept a `ViewRef` as in Input?

TimLariviere:
>What do you think about exposing this function as a event and accept a ViewRef as in Input?

Could you explain a bit more why you see it as an event?
By I agree that `ViewRef` seems our best option, even though I'm not a fan...

SergejDK:
@TimLariviere 
did it as an property - don't know why I was thinking about events ....

One thing that I caught was that using a `ViewRef` for `CarouselView` needs a `let carouselViewRef = ViewRef<CustomCarouselView>` instead of `ViewRef<CarouselView>`. Maybe we should look at this ?

I think I need to rewrite the `updateIndicatorViewRef` function so it is possible to have params like `prevValueOpt: ViewRef<'T> voption` instead of a specific type. - What do you think?

SergejDK:
@TimLariviere 
SwipeView is working fine and IndicatorView, too :) Samples are included in AllControls.

SergejDK:
@TimLariviere 
do you have a better solution for `CarouselVerticalItemsLayout` than I currently have?

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -675,6 +675,13 @@ module ViewUpdaters =
             let realTarget = (target :> Xamarin.Forms.IShellContentController).Page
             if realTarget <> null then currValue.UpdateIncremental(prevValue, realTarget)            
         | ValueSome _, ValueNone -> target.ContentTemplate <- null
+
+    let updateShellNavBarHasShadow prevValueOpt currValueOpt target =
+        match prevValueOpt, currValueOpt with
+        | ValueSome prevValue, ValueSome currValue when prevValue = currValue -> ()
+        | ValueNone, ValueNone -> ()
+        | _, ValueSome currValue -> Shell.SetNavBarHasShadow(target, currValue)
+        | ValueSome _, ValueNone -> Shell.SetNavBarHasShadow(target, true)
```
TimLariviere: The default value of `NavBarHasShadow` is dependent of the platform. If it's Android, it is `true`, `false` otherwise.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -723,4 +730,20 @@ module ViewUpdaters =
         | ValueSome prevValue, ValueSome currValue when prevValue = currValue -> ()
         | ValueNone, ValueNone -> ()
         | _, ValueSome currValue -> Element.SetMenu(target, currValue.Create() :?> Menu)
-        | ValueSome _, ValueNone -> Element.SetMenu(target, null)
\ No newline at end of file
+        | ValueSome _, ValueNone -> Element.SetMenu(target, null)
+
+    let updateIndicatorViewRef (prevValueOpt: ViewRef<CustomCarouselView> voption) (currValueOpt: ViewRef<CustomCarouselView> voption) (target: Xamarin.Forms.IndicatorView) =
```
TimLariviere: Name should contain the property updated. So `updateIndicatorViewItemsSourceBy`.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -723,4 +730,20 @@ module ViewUpdaters =
...
+        match prevValueOpt, currValueOpt with
+        | ValueSome prevValue, ValueSome currValue when prevValue = currValue -> ()
+        | ValueNone, ValueNone -> ()
+        | _, ValueSome currValue -> 
+            match currValue.TryValue with
+            | Some v -> IndicatorView.SetItemsSourceBy(target, v)
+            | None -> ()
```
TimLariviere: Maybe we should set the ItemsSource to `null` in the case the ref is invalid, instead of doing nothing and keeping a previous state no longer wanted?

Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json
```diff
@@ -624,6 +624,13 @@
           "defaultValue": "Xamarin.Forms.Color.Default",
           "inputType": "Xamarin.Forms.Color",
           "updateCode": "ViewUpdaters.updateShellUnselectedColor"
+        },
+        {
+          "source": null,
+          "name": "ShellNavBarHasShadow",
+          "defaultValue": "true",
```
TimLariviere: Same as above. Default value is dependent of the platform.

TimLariviere:
>do you have a better solution for CarouselVerticalItemsLayout than I currently have?
@SergejDK There doesn't seem to be a better way. Relying directly on `LinearItemsLayout.CarouselLayout` would lead to accidental overriding on the value since it's a read-only class.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewHelpers.fs
```diff
@@ -168,3 +168,9 @@ module ViewHelpers =
                 let res = ViewElement(externalObj.GetType(), create, update, attribs)
                 externalsTable.Add(externalObj, res)
                 res
+
+    // Helper for checking Platform - especially ShellHasNavBarShadow
+    let updateShellNavBarHasShadowHelper =
```
TimLariviere: I would directly write the "match ... with" instruction inside `Shell.SetNavBarHasShadow` on line 684 of ViewUpdaters.fs
You can ignore the `defaultValue` field in the bindings file since the `updateCode` field is provided.

SergejDK: Okay didn't know that the defaultValue is not needed if updateCode is there. will change it right now

TimLariviere: Oh. Apparently, I coded CodeGen to require the field even if unused.
Should fix that later. For now, only put an empty value for `defaultValue`.

Also I've noticed a few warnings from CodeGen

>Property 'IndicatorLayout' on type 'IndicatorView' not found
Property 'IndicatorShape' on type 'IndicatorView' not found
Property 'Item' on type 'SwipeItems' not found

SergejDK: I can understand why `item`may not be found: 
`public virtual ISwipeItem Item { get; set; }`

Don't know about the Indicatorproperties. Looking good to me or maybe I need to specify the type? https://docs.microsoft.com/en-us/xamarin/xamarin-forms/release-notes/4.4/4.4.0-api

TimLariviere: `IndicatorShape` is missing an 's'. It's `IndicatorsShape`.

`IndicatorLayout` bindable property is not marked as public. Instead, it appears we need to use `IndicatorTemplate`. So it can be removed from the binding.
https://github.com/xamarin/Xamarin.Forms/blob/bbb07f6f0ff533299a6fa183bc86e6be88ceecb0/Xamarin.Forms.Core/IndicatorView.cs#L36

And for some reasons, `Item` of SwipeItems is on the release notes, but not in the source code...

SergejDK: In jsut removed those and thanks for the hint of the missing s :D

magnushammar:
Awesome work thanks!

I just compiled an app with it now and my SwipeItems are showing as expected.

PR #673, opened 2020-02-19
Enable automatic builds on v*.* branches & release on Git Tag

TimLariviere:
Like discussed in https://github.com/fsprojects/Fabulous/pull/669#issue-374422035

Only master is not automatically releasing to MyGet

### DIFF
--- a/azure-pipelines-full.yml
+++ b/azure-pipelines-full.yml
@@ -1,5 +1,6 @@
 trigger:
 - master
+- v*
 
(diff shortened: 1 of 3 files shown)


TimLariviere:
@SergejDK Ok, so GitHub Release randomly failed to upload artifacts so I deactivated artifacts upload.
And now, it works!
Just pushed 0.52.0 to NuGet by creating the tag :)

PR #677, opened 2020-02-23
More modular AllControls sample

TimLariviere:
I'm reworking the AllControls sample to make it easier to read a specific sample (e.g. How to use SkiaSharp?) and as well as make it easier to add new samples to it.

The idea is to add have distinct files and register them in the Samples.fs file, so the app will automatically pick them up.

Currently moved SkiaSharp, OxyPlot and Maps

@SergejDK Let me know what you think of it

Screen-Recording-2020-02-23-at-7


SergejDK:
@TimLariviere 
I think this is a great idea. The current AllControls-sample was getting way to big to handle and no one could find a specific implementation easily. This even shows how to build bigger applications with more screens :)

I'll take a look on what exactly you did in the next days.

TimLariviere:
Finished moving samples away from the big file. :)
Tested it on iOS for now. Other platforms might not build because of dependencies to extensions.

SergejDK:
@TimLariviere 
Works amazing and pretty good examples especially for the extensions. Code looks good too even though I needed a bit to understand it :)

Tested on:
- iOS -> works good
- android -> currently my setup is kinda broke need to check this before i can test it...
- macOs -> 
    - ScrollView `Scroll`-event not working
    - Swipeview / Refreshview not working but I think mac is not supported
    - Can't get back from navigationpage
- GTK / WPF / UWP -> Could not test because of my current setup

Maybe one more thing - we should show the control in die videomanagerextensions in this example else the sound keeps on playing which is kinda annoying.

TimLariviere:
@SergejDK Thanks for the feedback!

>Code looks good too even though I needed a bit to understand it :)

Do you have any suggestion how we can make this part easier to understand?

>macOS: Can't get back from navigationpage

Yes, I noticed this as well. I wasn't expecting Android/iOS to allow having a Shell/NavigationPage inside another NavigationPage but it worked. Though macOS doesn't allow this.
Will need to add the famous "Go back" button manually for macOS only.

>Maybe one more thing - we should show the control in die videomanagerextensions in this example else the sound keeps on playing which is kinda annoying.

Noticed that as well.
For some reasons, the video player doesn't get disposed when navigating back.
Need to investigate this.

SergejDK:
> Do you have any suggestion how we can make this part easier to understand?

I just think that a little code documentation is enough - just did not worked a lot with f# the last weeks.

> Yes, I noticed this as well. I wasn't expecting Android/iOS to allow having a Shell/NavigationPage inside another NavigationPage but it worked. Though macOS doesn't allow this.
Will need to add the famous "Go back" button manually for macOS only.

This should not be to difficult - maybe seperate issue `good first issue` ?

Issue #682, opened 2020-03-01
Xamarin.Forms 4.5


Liminiens:
https://devblogs.microsoft.com/xamarin/xamarin-forms-4-5/

It would be nice to have a MediaElement wrapper :)

Happypig375:
Given that [it took 2 months for Fabulous to update to Xamarin Forms 4.4,](https://github.com/fsprojects/Fabulous/pull/644) you will most probably need to wait a similar amount of time to be able to use this.

TimLariviere:
One can speed the availability of support for XF 4.5 by sending a PR ;)

Here's all the steps for such a PR:
- Update the used version of Xamarin.Forms for [NuGet](https://github.com/fsprojects/Fabulous/blob/master/Packages.targets) and [Paket](https://github.com/fsprojects/Fabulous/blob/master/paket.dependencies) (lowest major version only - eg. 4.5.0.356)
- Update the [JSON file](https://github.com/fsprojects/Fabulous/blob/master/Fabulous.XamarinForms/src/Fabulous.XamarinForms/Xamarin.Forms.Core.json) that generates the wrappers according to the [API changes of XF 4.5](https://docs.microsoft.com/en-us/xamarin/xamarin-forms/release-notes/4.5/4.5.0-api)
- Add samples for the new controls/properties in the [AllControls project](https://github.com/fsprojects/Fabulous/tree/master/Fabulous.XamarinForms/samples/AllControls)
- Write any appropriate [unit tests](https://github.com/fsprojects/Fabulous/tree/master/Fabulous.XamarinForms/tests/Fabulous.XamarinForms.Tests)
- Write any appropriate [documentation](https://github.com/fsprojects/Fabulous/tree/master/docs/Fabulous.XamarinForms)

Even if just the 2 first points are done and a PR is submitted, this will be an immense help for us. :)

vshapenko:
@TimLariviere, are there any thoughts about simplifiying the process? Currently it looks routine and boring, so would be great to have a kind of guidance (from you, as a maintainer), how to simplify it. Maybe some code needs to be rewritten, some concepts need rethink. I am ready to put my resources on such tasks(cause this is important for the future of project i am developing currently), but i need a guidance.

TimLariviere:
@vshapenko Unfortunately, there's only so much you can simplify or automate.

>Maybe some code needs to be rewritten, some concepts need rethink

[...]

PR #700, opened 2020-03-28
Reference work on XF 4.5

TimLariviere:


Happypig375:
Fixes #682

TimLariviere:
IndicatorView will be a PITA.
They change it from having a CarouselView property to CarouselView having a IndicatorView property, though the value is still stored in the IndicatorView instance.

It means we won't be able to clear the value easily, or even at all...

SergejDK:
@TimLariviere 
do you think we need to handle those methods `OnChildRemoved `

TimLariviere:
@SergejDK No, I don't think we need to. It's not a user-triggered event so it's not really useful in Fabulous.

TimLariviere:
Replaced by #709

PR #701, opened 2020-03-29
XF Upgrade

SergejDK:
@TimLariviere 

you did most of the work ;)

I just made simple changes but I think you can take a look on this.

Currently I have a problem: indicatorView is not shown as a property of carouselview ?

Wanted to include the samples...

Waht do you think about the triggers?

https://devblogs.microsoft.com/xamarin/xamarin-forms-4-5/


TimLariviere:
https://github.com/SergejDK/Fabulous/blob/8d91a4ddfa1dff73572b2a5c932dd557f980b968/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs#L759

Noticed this property has in fact a bindable property so we're supposed to let CodeGen handle the update code.
Though for some reason, it's marked private so we're force to use this helper function for the moment.

I opened an issue on Xamarin.Forms so they make the field public.
https://github.com/xamarin/Xamarin.Forms/issues/10148

TimLariviere:
About `ImageSource` and `MediaSource`, I put them in the json file though I'm not sure if we should go that way.

I did that because I noticed they inherit from `Xamarin.Forms.Element` like the other controls.

But there's no added value to write
`View.Image(Source = View.UriImageSource(Uri = "path/to/image.png"))`
instead of
`View.Image(Source = Path "path/to/image.png")`

I think we need to remove all ImageSource and MediaSource derived types from the json file and add the new cases to the [InputTypes](https://github.com/SergejDK/Fabulous/blob/8d91a4ddfa1dff73572b2a5c932dd557f980b968/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/InputTypes.fs#L6-L13)

TimLariviere:
>Currently I have a problem: indicatorView is not shown as a property of carouselview ?

@SergejDK Yes. That's the issue with IndicatorView/CarouselView.
In XF 4.4, IndicatorView had a CarouselView property (with matching bindable property field).
In XF 4.5, they removed the property from IndicatorView (but left the bindable property field) and instead put a "fake" property on CarouselView which only put the value on the IndicatorView instance...
So CodeGen is not able to detect that, we need to write our own `updateCode` and set the property to `"source": null` in the json file to let CodeGen knows we will handle it.

SergejDK:
> I opened an issue on Xamarin.Forms so they make the field public.

The only reason why I wrote it. 

> I think we need to remove all ImageSource and MediaSource derived types from the json file and add the new cases to the InputTypes

Was thinking about this, too. 

> Also could you write samples in AllControls to make sure the new controls work and are usable, please?

Will do this.


> So CodeGen is not able to detect that, we need to write our own updateCode and set the property to "source": null in the json file to let CodeGen knows we will handle it.

Interesting stuff which they do there :D

TimLariviere:
>What do you think about the triggers?

@SergejDK That's a good question. We don't currently have any support for them in Fabulous...
I think we will need to open an issue about them and brainstorm how we should support them.

SergejDK:
> I think we need to remove all ImageSource and MediaSource derived types from the json file and add the new cases to the InputTypes

I think it would be better to split between Image and Media because it seems like Media is sth. like a video?

TimLariviere:
@SergejDK Yes, of course. Media are videos and need a separate DU.

SergejDK:
@TimLariviere 
Changed some parts - still need to add some example. I think the most important part is the MediaElement

SergejDK:
@TimLariviere 
I am ready with those parts.

TimLariviere:
Replaced by #709

PR #705, opened 2020-04-11
Add a sentence to clarify the documentation structure

x-ji:


x-ji:
I opened the page several times thinking that this is the whole of the documentation and the links at the beginning of the page are just for different parts of the page... An additional clarifying sentence would help avoid confusion IMO.

PR #707, opened 2020-04-14
Enhanced FabulousWeather sample

TimLariviere:
Reworked the FabulousWeather sample for a better UI, use of some of the latest Xamarin.Forms controls (CarouselView, RefreshView, etc), some platform-specific logic, and display of hourly forecast.

enhanced-fabulous-weather


TimLariviere:
The code could use some clean up still

PR #720, opened 2020-04-24
Make path to follow clearer

dsyme:
The Fabulous home page is unclear

* The main "Fabulous" link takes you to unfinished docs

* The "StaticView" is not really supported 

* The "CodeGen" is a tool and is not documented

We really have to make the getting-started path really simple.   I think we should just direct people down the happy path for now and take them straight to the "Fabulous for Xamarin Forms" documentation.  We really don't want anyone to end up anywhere else for now??

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,12 +6,6 @@ Never write a ViewModel class again! Conquer the world with clean dynamic UIs!
 
 * [Documentation of Fabulous](https://fsprojects.github.io/Fabulous/Fabulous.XamarinForms/)
 
-Tools
------
-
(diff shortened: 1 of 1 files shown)


dsyme:
@TimLariviere I'm going to merge this as we should get this clear before the show goes live

TimLariviere:
@dsyme I'm fine with it, but I think we should still use the name Fabulous.XamarinForms to match READMEs and existing docs.

Even though we might commonly call Fabulous.XamarinForms Fabulous, there's a difference between the two - notably the NuGet packages.

PR #721, opened 2020-04-24
WIP: Concept of key property for ViewElement

### LINKED ISSUE #713: Renderer is recreated on view update
1. Run this example on Android 7.0+: [SqueakyApp.zip](https://github.com/fsprojects/Fabulous/files/4508915/SqueakyApp.zip)
2. Increment counter multiple times, or use Timer switch
3. See Device Log:
```
Time	Device Name	Type	PID	Tag	Message
04-21 12:09:04.880	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:06.807	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:06.772	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:06.564	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:06.297	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:06.045	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:05.793	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:05.569	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:05.336	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:05.112	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:04.689	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:04.649	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:04.425	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:04.195	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:04.016	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.964	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.737	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.504	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.311	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.274	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:03.042	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.808	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.619	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.577	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.390	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.345	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:02.119	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:01.867	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:01.574	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:01.308	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:01.056	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:00.786	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:00.534	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:00.266	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:09:00.011	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:59.747	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:59.495	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:59.226	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:58.977	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:58.691	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:58.449	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:58.191	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:57.938	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:57.679	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:57.436	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:57.177	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:56.930	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
04-21 12:08:52.979	Google Pixel 3	Info	28526	mono-stdout	Create RtcVideoFrameRenderer
```

The problem occurs when dynamically adding and deleting any element inside the grid (in this case). It seems that all elements in grid disposing and recreating even if they were not deleted by condition. I found workaround for this bug. You can set `fixBug` value to `true` and the problem will go away. As you can see in code below I just change element visibility with `fixBug` value what does not delete element itself.

```fsharp

let view (model: Model) dispatch =
    let vis = model.Count % 2 = 0
    let fixBug = false

    View.ContentPage(
        content = View.StackLayout(padding = Thickness 20.0, verticalOptions = LayoutOptions.Center,
        children = [
            View.Label(text = sprintf "%d" model.Count, horizontalOptions = LayoutOptions.Center, width=200.0, horizontalTextAlignment=TextAlignment.Center)
            View.Button(text = "Increment", command = (fun () -> dispatch Increment), horizontalOptions = LayoutOptions.Center)
            View.Button(text = "Decrement", command = (fun () -> dispatch Decrement), horizontalOptions = LayoutOptions.Center)
            View.Label(text = "Timer", horizontalOptions = LayoutOptions.Center)
            View.Switch(isToggled = model.TimerOn, toggled = (fun on -> dispatch (TimerToggled on.Value)), horizontalOptions = LayoutOptions.Center)
            View.Slider(minimumMaximum = (0.0, 10.0), value = double model.Step, valueChanged = (fun args -> dispatch (SetStep (int (args.NewValue + 0.5)))), horizontalOptions = LayoutOptions.FillAndExpand)
            View.Label(text = sprintf "Step size: %d" model.Step, horizontalOptions = LayoutOptions.Center) 
                
            View.Grid(
                padding = Thickness(0.),
                rowSpacing = 0.,
                rowdefs = [
                    Dimension.Star
                    Dimension.Absolute 64.
                    Dimension.Auto
                    Dimension.Auto
                    Dimension.Star
                    Dimension.Absolute 64.
                    Dimension.Absolute 20.
                ],
                children = [

                    if fixBug then
                        View.AbsoluteLayout(
                        horizontalOptions = LayoutOptions.Fill,
                        verticalOptions = LayoutOptions.Fill,
                        isVisible = vis,
                        children = [
                        ]
                        ) |> rowSpan 6
                      
                    if not fixBug && vis then
                        View.AbsoluteLayout(
                          horizontalOptions = LayoutOptions.Fill,
                          verticalOptions = LayoutOptions.Fill,
                          children = [
                          ]
                        ) |> rowSpan 6

                    View.RtcVideoFrame(
                      automationId = "RemoteVideoFrame",
                      margin = Thickness(0.),
                      ref = model.VideoRef,
                      horizontalOptions = LayoutOptions.FillAndExpand,
                      verticalOptions = LayoutOptions.FillAndExpand,
                      isVisible = vis
                    ) |> rowSpan 6
                ]
                )

            if model.Count % 2 = 0 then
                View.Button(text = "Reset", horizontalOptions = LayoutOptions.Center, command = (fun () -> dispatch Reset), commandCanExecute = (model <> initModel))
        ]))
```

vshapenko:
@TimLariviere , could you please tell me if i go right way in my attemps to fix #713 ?


Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -38,31 +38,44 @@ module ViewUpdaters =
...
+                    
                 // Adjust the existing targetColl and create the new targetColl
                 for i in 0 .. coll.Length-1 do
                     let newChild = coll.[i]
+                    let key = getKey newChild
                     let prevChildOpt = match prevCollOpt with ValueNone -> ValueNone | ValueSome coll when i < n -> ValueSome coll.[i] | _ -> ValueNone
                     let prevChildOpt, targetChild = 
                         if (match prevChildOpt with ValueNone -> true | ValueSome prevChild -> not (identical prevChild newChild)) then
-                            let mustCreate = (i >= n || match prevChildOpt with ValueNone -> true | ValueSome prevChild -> not (canReuse prevChild newChild))
+                            let mustCreate = ((not <| targetControls.ContainsKey(key)) || match prevChildOpt with ValueNone -> true | ValueSome prevChild -> not (canReuse prevChild newChild))
```
TimLariviere, line 78: This coupled with the fact you generate a random key for all controls effectively prevents reuse for controls with no user-defined key.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -38,31 +38,44 @@ module ViewUpdaters =
...
                 let _ = targetColl.Count
                 let n = targetColl.Count
-
+                let targetControls = Dictionary<string,'TargetT>()
+                for i in 0..prevColl.Length-1 do
+                        let key = getKey prevColl.[i]
+                        targetControls.[key]<-targetColl.[i]
+                         
+                while (targetColl.Count > coll.Length) do
+                  targetColl.RemoveAt (targetColl.Count - 1)
```
TimLariviere, line 68: Doing that, there's a chance you're deleting controls you might reuse.

Example:
```fs
// Previous
[
    View.Label(text = "Label")
    View.Button(key = "MyButton", text = "Button")
]

// Current
[
    View.Button(key = "MyButton", text = "SuperButton")
]
```

This will result in removing the button since it's outside of the new range.
But then you'll try to update it since the key is matching, even though it disappeared from the screen.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: I am agree this condition need to be more complex. Right now we have a problem - how to determine, which control we should cache, and which we should recreate. Unfortunately, we can not instantiate two "instance" of same view element on different view cycles.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: This will not happen, because i have separate dictionary for such controls. targetColl is not a place, where we seek for controls after we put them in dictionary.

src/Fabulous/ViewElement.fs
```diff
@@ -116,6 +117,11 @@ type ViewElement internal (targetType: Type, create: (unit -> obj), update: (Vie
 
     [<System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)>]
     static member val _CreatedAttribKey : AttributeKey<obj -> unit> = AttributeKey<_>("Created")
+   
+    [<System.ComponentModel.EditorBrowsable(System.ComponentModel.EditorBrowsableState.Never)>]
+    static member val _KeyAttribKey : AttributeKey<string> = AttributeKey<_>("Key")
```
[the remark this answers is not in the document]
vshapenko: I will take a look, thank you

src/Fabulous/ViewElement.fs
```diff
@@ -155,6 +164,15 @@ type ViewElement internal (targetType: Type, create: (unit -> obj), update: (Vie
 
     /// Differentially update the inherited attributes of a visual element given the previous settings
     member x.UpdateInherited(prevOpt: ViewElement voption, curr: ViewElement, target: obj) = update prevOpt curr target
+    
+    member x.KeyValue
+            with get() = key
+            and set v = key<-v
+    member x.Key _ =x
```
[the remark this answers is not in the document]
vshapenko: That is a requirement of code gen, we cannot put a attribute on ViewElement, only on Control itself, so we should "mock" the behavior for code to be compiled. Key  should be immutable, yes. i'll fix it

src/Fabulous/ViewElement.fs
```diff
@@ -108,6 +108,7 @@ type ViewRef<'T when 'T : not struct>() =
 /// A description of a visual element
 type ViewElement internal (targetType: Type, create: (unit -> obj), update: (ViewElement voption -> ViewElement -> obj -> unit), attribs: KeyValuePair<int,obj>[]) = 
     
+    let mutable key = Guid.NewGuid().ToString("N")
```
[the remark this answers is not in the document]
vshapenko: Yes, you are right, i'll correct this behaviour

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: i>=n is not always valid for support old style in new paradigm. It would be more likely like this: 
 (i>=n && (not <| targetControls.Contains key))

vshapenko:
@TimLariviere , could you please take another look at PR? I think i have resolved issues you mentioned

src/Fabulous/ViewElement.fs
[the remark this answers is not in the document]
TimLariviere: Since `Key` has been declared as `canBeUpdated: false` in the JSON file, CodeGen won't attempt to apply the value anywhere.
It will just only keep it in the `attribs` array passed as a constructor parameter.

So `x.Key _ = x` has no use and can be safely removed.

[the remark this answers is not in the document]
vshapenko: Oh, my bad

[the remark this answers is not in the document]
TimLariviere: x.KeyValue was deleted but not x.Key?

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
TimLariviere: It was an issue previously because the controls in excess were removed from the list Xamarin.Forms was using (e.g. ListView's `Items` or StackLayout's `Children`).
Even though they were kept in an another list so they could be reused, they were not reinserted to the Xamarin.Forms's list, removing them from the UI incorrectly.

But now, it's no longer an issue since you set all items an index.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
TimLariviere: I agree. The case where the control has no matching key or has no key should be handled differently. We shouldn't rely on index anymore in my opinion.
See my notes in the first message.

Fabulous.CodeGen/src/Fabulous.CodeGen/Generator/CodeGenerator.fs
```diff
@@ -243,6 +244,7 @@ module CodeGenerator =
                     match m.Name with
                     | "created" -> sprintf "%s?%s: (%s -> unit)" commaSpace m.Name data.FullName
                     | "ref" ->     sprintf "%s?%s: ViewRef<%s>" commaSpace m.Name data.FullName
+                    | "key" ->     sprintf "%s?%s: %s" commaSpace m.Name m.InputType
```
TimLariviere, line 247: There's a bunch of other checks like those 2 in the same file that should also have a `key` case.

src/Fabulous/ViewElement.fs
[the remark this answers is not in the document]
vshapenko: Removed, it is unused now, you are right

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: For now i am following this way: 
1. I check if  i have control in dictionary. If yes, i apply new logic. If no, old logic takes place

src/Fabulous/ViewElement.fs
```diff
@@ -146,6 +152,13 @@ type ViewElement internal (targetType: Type, create: (unit -> obj), update: (Vie
         match attribs |> Array.tryFind (fun kvp -> kvp.Key = key.KeyValue) with 
         | Some kvp -> unbox<'T>(kvp.Value)
         | None -> failwithf "Property '%s' does not exist on %s" key.Name x.TargetType.Name
+        
+    member x.TryGetKey() =
+        match x.TryGetAttributeKeyed(ViewElement._KeyAttribKey) with
+        | ValueSome key -> Some key
+        | ValueNone -> None
```
TimLariviere, line 159: Please use `ValueOption` instead of `Option` when working on the internal plumbing of Fabulous.
ValueOption is a struct and is lighter on the memory (stored on Stack instead of the GC-managed Heap).
So I think you can directly return the result of TryGetAttributeKeyed

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -38,6 +38,7 @@ module ViewUpdaters =
            (create: 'T -> 'TargetT)
            (attach: 'T voption -> 'T -> 'TargetT -> unit) // adjust attached properties
            (canReuse : 'T -> 'T -> bool) // Used to check if reuse is possible
+           (getKey:'T->string option)
```
TimLariviere, line 41: Same here. Please use `ValueOption`.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
TimLariviere: I fear that the old logic can't really be kept as-is with randomly placed keyed controls.
The previous logic was only working since all controls were check based on the index indiscriminately.

TimLariviere:
One thing that will greatly help for this PR is unit testing.
Could you add tests in the `Fabulous.XamarinForms.Tests` projects that only call the `updateCollectionGeneric`, please?

With a bunch a realistic cases such as:
- No keyed controls & no change between 2 calls to `updateCollectionGeneric`
- No keyed controls & remove control at end of list between 2 calls to `updateCollectionGeneric` (should work today)
- No keyed controls & remove control at start of list between 2 calls to `updateCollectionGeneric` (should not work correctly)
- 2nd control with a key & remove 1st one at start of list between 2 calls to `updateCollectionGeneric` (should work correctly with the new algorithm)

Example of unit test (not actual code)
```
let previous =
    [
        View.Label(text = "Label 1")
        View.Label(text = "Label 2")
    ]

let current = 
    [
        View.Label(text = "Label 1")
        View.Label(text = "Label 2")
    ]

let mutable mockUpdateCallCount = 0
let mockUpdate _ _ =
    mockUpdateCallCount <- mockUpdateCallCount + 1

do updateCollectionGeneric previous current ... mockUpdate ...

mockCreateCallCount |> should equal 0
mockUpdateCallCount |> should equal 2
// And so on
```

With those tests, we'll be able to tell if the new algorithm correctly works, as well confirmed previous codes still work as expected.

Fabulous.CodeGen/src/Fabulous.CodeGen/Generator/CodeGenerator.fs
vshapenko: Ok

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok

vshapenko:
@TimLariviere , do i need to create  actual XF controls in tests? Cause i need somehow to mock create function

TimLariviere:
>do i need to create actual XF controls in tests? Cause i need somehow to mock create function

No. By mocking the create function, we can ignore Xamarin.Forms.
We don't want to test if the create function does its work, but if it's called when expected.

vshapenko:
Ok. i have made a little bunch of tests. Looks like if we need more, would be good to use something like FsCheck

TimLariviere:
@vshapenko I will add in this PR another file with a set of non-regression tests and new use cases that should be supported by the new algorithm.
We will be able to fix the code after that.

TimLariviere:
And I must say I don't have much experience using FsCheck, so I don't really see how we can use it here.
So you have a specific idea, don't hesitate to try it out here.

vshapenko:
Maybe there is no need in FsCheck yet.  I do not yet fully understand the conditions that should be met to mark test passed on different set of views transformations and updates.

TimLariviere:
@vshapenko I've merged the non-regression tests from #725 into this PR.
The tests using the new key attribute are missing, I will add them here.
To start, we will need to make the existing tests pass.

vshapenko:
Existing tests look ok

vshapenko:
@TimLariviere , what else do we need to safely merge this PR?

TimLariviere:
@vshapenko I wrote some new tests making use of key. Some are failing.
I'll need to clean up the tests a bit and I will push so we can fix the algorithm.

TimLariviere:
There's 6 tests failing in the ones I've just added (keyed and mixed use cases).
Also I figured out that we can reuse elements directly like the keyed ones, if the instance is the same (`identical`).

I had to rewrite the algorithm a bit to facilitate its testing.
Now `updateCollectionGeneric` became `updateCollectionGenericInternal` and don't directly interact with a list of Xamarin.Forms controls but instead gives the operations to move from the previous state to the new one.
E.g.

```
[
    View.Label(text = "First time")
    View.Label(key = "Title", text = "Welcome")
    View.Button(text = "Click here")
]
-->
[
    View.Label(key = "Title", text = "Welcome")
    View.Button(text = "Click here")
]
```

Results in

```
Update the previous label keyed Title with the new label keyed Title
Update the previous button with the new button (since it has a compatible type)
Remove the label "First time" (since it couldn't be reused)
```

`updateCollectionGeneric` applies the operations to the actual Xamarin.Forms list, so it doesn't change its usage.

I hope the tests are understandable with their naming convention.

Don't hesitate to ask questions.

vshapenko:
@TimLariviere , looks like your version of algorithm tries to store previous view elements. This would not work this way, i think. My vision is that we operate keys and tie keys to 'TTarget rather than 'T. Otherwise, it will bring alot of inconsistent behaviour, as we see in tests.

TimLariviere:
I don't think storing `'T` instead `'TTarget` should make a difference.

In Fabulous, a necessary condition is that the current UI (`'TTarget`) should always match the previous state (`'T`) so when selecting a `'T` we can find the associated `'TTarget` (they're using the same index in `prevColl` / `targetColl`).

So we can abstract the real controls list and only work with the previous/current ViewElements.

TimLariviere:
Note: I had to keep a copy of the `targetColl` to avoid accessing an already updated list.
```
let previousTargetColl = List(targetColl)
```

Though I believe we can achieve the same without this copy.

vshapenko:
I think, currently root of all evils is at  following code: 

``` let prevChildOpt = match prevCollOpt with ValueNone -> ValueNone | ValueSome coll when i < n -> ValueSome coll.[i] | _ -> ValueNone```

**Comment by TimLariviere** [comment]
Yes, we should not use index anymore.

How I imagine it:
We make 2 sets from the previous collection: one with all the keyed elements, one with the rest ("the reusable pool").

For each element (`curr`) inside the new collection:
- if prevColl has an element where `identical prev curr = true`
    - move to the correct index but do not update (ViewElement is immutable so same ref = same values)
- else if `curr` has a key
    - try to get it from the keyed list
        - update it
    - otherwise try to get an element from the reusable pool where `canReuse prev curr = true`
        - update it (then remove the reused element from the pool)
    - else
        - create it
- else
    - try to get an element from the reusable pool where `canReuse prev curr = true`
        - update it (then remove the reused element from the pool)
    - else
        - create it

**Comment by vshapenko** [comment]
@TimLariviere , there are some strange cases in tests like 
Given previous state = 1k / current state = 2k, updateCollectionGeneric should Update[1k -> 2k]
For test to pass, they should have same automationId, i think, else  we dont get canReuse=true

**Comment by TimLariviere** [comment]
That's a tricky case, and maybe it's wishful thinking, but `1k` has no matching counterpart so it can be "demoted" into the reusable pool.
And since `2k` is a brand new key, it can't find matching keys in the keyed list, so it will tap into the reusable pool, finding `1k`.

One way I can think of handling that is: before adding a previous keyed element inside the keyed list, we check that a matching key exists in the new collection. if exists, add to keyed list, if not, add to reusable pool.

Might not be the most performant way but we'll see how to improve once we cover all the cases.

File: Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewHelpers.fs [hunk 9]
```diff
@@ -14,11 +14,13 @@ module ViewHelpers =
     let identical (x: 'T) (y:'T) = System.Object.ReferenceEquals(x, y)
     
     let getKey (el:ViewElement) = el.TryGetKey()
-          
+    
+    let sameType (prevChild: ViewElement) (newChild: ViewElement) =
+        prevChild.TargetType = newChild.TargetType
```
TimLariviere, line 19: Don't add this function, use `canReuse` instead.
`canReuse` already does this check along with other mandatory checks to make sure we can really update a control.

vshapenko: Ok, so testcases need to be updated at least same automationId should be set

TimLariviere: Which tests?
There should be no need for automationId in my opinion.

vshapenko: canReuse relies on automationId in its checks

TimLariviere: Yes, it only checks that the AutomationId didn't changed since XF prohibits changing an already set AutomationId (you can go from nothing to a value, but not from a value to another value).

So since the tests never define an AutomationId, it won't influence the result of canReuse (if it's returning false, it's another reason).

vshapenko: Yes, you are right

vshapenko:
Ok, for now i have only 4 tests failing. Most cases are about "replace". In our logic it should be create& remove i believe

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+            (getKey: 'T -> string voption)
+            (canReuse: 'T -> 'T -> bool)
+            (move: int -> 'T -> unit)
+            (create: int -> 'T -> unit)
+            (update: int -> 'T -> 'T -> unit)
+            (remove: int -> unit)
+            (clear: unit -> unit) =
+        
+        match prevCollOpt, collOpt with 
+        | ValueSome prevColl, ValueSome newColl when identical prevColl newColl -> ()
```
TimLariviere, line 50: Just thought of another case where we can do nothing here: ValueNone, ValueNone

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+                 let key = getKey newChild
+                 match key with
+                 | ValueSome key -> newKeys.Add key |> ignore
+                 | ValueNone _ -> ()
+                 
+            for i in 0..n-1 do
+                let key = getKey prevColl.[i]
+                match key with
+                | ValueSome key ->
+                    availableKeyedElements.[key] <- prevColl.[i]
```
TimLariviere, line 79: I would check if the key is also in `coll` here.
If it's there then add to `availableKeyedElements` otherwise add to `rest`.
It allows you to not recheck it later.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+            // Adjust the existing targetColl and create the new targetColl
+            for i in 0 .. coll.Length-1 do
+                let newChild = coll.[i]
+                let key = getKey newChild
+                let elems =
+                        seq{
+                            yield! rest
+                            yield! availableKeyedElements.Values
+                           }
+                if (prevColl|>Seq.exists (identical newChild)) then
```
TimLariviere, line 101: Nitpick: `prevColl` is known to be an array, so I think it's clearer to use `Array.exists` instead of `Seq.exists` :D

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+                    move i newChild
+                    match getKey newChild with
+                     | ValueSome key -> availableKeyedElements.Remove key |> ignore
+                     |_ ->rest.Remove newChild |> ignore
```
TimLariviere, line 105: Could you put this in a function and reuse it where needed, please?

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+                     
+                elif (key.IsSome)
+                then
+                  if(contains (key)) then 
+                     let previousKeyedElement = availableKeyedElements.[key.Value] // We should be safe here
+                     if(canReuse previousKeyedElement newChild) then
+                      update i previousKeyedElement newChild
+                      availableKeyedElements.Remove key.Value |> ignore
```
TimLariviere, line 113: You're missing an `else` case here.
What happens if the element has a matching key, but it is not reusable?

e.g.
```
View.Label(key = "MyKey")
->
View.Button(key = "MyKey")
```

TimLariviere: See test `Given previous state = Txk / current state = Tyk (same key), updateCollectionGeneric should replace Txk with Create[Tyk]`

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+                  else
+                     match (elems|>Seq.tryFind(fun c-> canReuse c newChild)) with
```
TimLariviere, line 115: You could directly search `rest` here instead of `elems`.
With the 2nd comment I made, `availableKeyedElements` will only contain elements reused inside `if contains key`.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+                     | Some el ->
+                         update i el newChild
+                         match getKey el with
+                         | ValueSome key -> availableKeyedElements.Remove key |> ignore
+                         |_ ->rest.Remove el |> ignore
+                     | None ->
+                        create i newChild
+                else
+                   
+                    match (elems|>Seq.tryFind(fun c-> canReuse c newChild && (not <| newKeyContains c))) with
```
TimLariviere, line 125: Same here. Search inside `rest` and not `elems`.
Also `not <| newKeyContains c` has no longer any meaning with the 2nd comment I made.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,120 @@ module ViewUpdaters =
...
+
+            let elems =
+                        seq{
+                            yield! rest
+                            yield! availableKeyedElements.Values
+                           }
+                        
+            for el in elems do
+                let index= prevColl|>Array.findIndex (fun c-> c=el)
+                remove index
```
TimLariviere, line 143: Same here. You could only remove elements still inside `rest`. All `availableKeyedElements` will have been reused at this point.

TimLariviere: Though on second thought, I think we need to change this part a bit.

Since we reused everything that was reusable and created the correct controls at the correct indexes, we are sure that for index 0 to index (coll.Length - 1) everything is correct.
What we really want here is to delete the eventual excess (from coll.Length to prevColl.Length - 1).

So we could only go with the previous while loop there was at the beginning of the PR.

But it makes it hard to unit test. In the tests, we want to know which elements did not make it (e.g. what's remaining in `rest` at the end).

So the only way I can think of handling this would be to change `remove: int -> unit` to `remove: (lastIndex: int) -> (nonReusedElements: ViewElement list)`.
`updateCollectionGeneric` would remove all items after the last index of `coll` (like we did before).
And the tests would check the removed elements like currently.

TimLariviere:
>Ok, for now i have only 4 tests failing. Most cases are about "replace". In our logic it should be create& remove i believe

Yes, I agree.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: OK

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
vshapenko: Ok, i will try this

vshapenko: Do i understand correclty, that we should delete all elements from targetColl that exceeds lastIndex?

TimLariviere: Yes

vshapenko:
@TimLariviere , please, take a look to changes, all tests are passing now, but i am in doubt about mockRemove and remove implementation

Fabulous.XamarinForms/tests/Fabulous.XamarinForms.Tests/ViewUpdaters/UpdateCollectionGenericTests.fs
```diff
@@ -7,113 +7,151 @@ open NUnit.Framework
...
+            let prevIndex = previousCollection.Value |> List.findIndex (fun c -> c = previousChild)
+            operations.Add(i, Update (prevIndex, previousChild, newChild))
             
-        let mockAttach previousViewElement currentViewElement _ =
-            operations.Add(Attach (previousViewElement, currentViewElement))
+        let mockRemove i unused=
+            let mutable offset = 0
+            for el in unused do
+             operations.Add(i + offset, Remove el)
+             offset<-offset+1
```
TimLariviere, line 40: Here the first parameter is just a sorting key for the tests, it's not directly linked to `i`.
For `Remove`, we can just go add things at the end (`offset` ensures we stay at the end of the operations list)

So this should work:
```
let mockRemove _ unused =
    let offset = match newCollection with ValueNone -> 0 | ValueSome coll -> coll.Length
    for i = 0 to unused.Count - 1 do
        operations.Add(offset + i, Remove unused.[i])
```

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -38,45 +147,60 @@ module ViewUpdaters =
...
+                | _ -> [||]
+                
+            for el in unused do
+                let idx = prevColl|> Array.findIndex (fun c-> c=el)
+                let target = previousTargetColl.[idx]
+                targetColl.Remove target |> ignore
+                
+            if(targetColl.Count>lastIndex) then
+                    for i=lastIndex to targetColl.Count-1 do
+                        targetColl.RemoveAt i
```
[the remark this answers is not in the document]
vshapenko: Ok , i think trim should be good

TimLariviere:
And you were right about the "replace" tests, it should be "create + remove".
I changed them.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/ViewUpdaters.fs
```diff
@@ -28,6 +28,115 @@ module ViewUpdaters =
...
+    /// If no reuse is possible, the element will create a new control.
+    let updateCollectionGenericInternal
+            (prevCollOpt: 'T[] voption)
+            (collOpt: 'T[] voption)
+            (getKey: 'T -> string voption)
+            (canReuse: 'T -> 'T -> bool)
+            (move: int -> 'T -> unit)
+            (create: int -> 'T -> unit)
+            (update: int -> 'T -> 'T -> unit)
+            (remove: int->'T seq -> unit)
```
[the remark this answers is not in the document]
vshapenko: > And you were right about the "replace" tests, it should be "create + remove".
> I changed them.

For current logic it looks like it is not needed, because we clear rest on create, and remove just does not take place because it is useless in such scenario. So, we make create and do not need remove because there is nothing to remove

vshapenko:
@TimLariviere , all tests are passing right now. I think we do not need "create+remove" logic in tests.

TimLariviere:
Looks good.

I cleaned up the code a little, adding comments and reformatting to follow same style than the rest of Fabulous.

I also added an `identicalElements` list so we don't need to remove an identical element (without a key) from reusableElements at the end.

>For current logic it looks like it is not needed, because we clear rest on create, and remove just does not take place because it is useless in such scenario. So, we make create and do not need remove because there is nothing to remove

Even though I agree with you that it doesn't change anything for the algorithm, I prefer to have create + remove instead of just create.
It makes the intent clearer. We can't reuse the previous element, so we create a new one and delete the previous one.

TimLariviere:
Tested the samples.
All work except TicTacToe. Will take a look later.

vshapenko:
> Looks good.
> 
> I cleaned up the code a little, adding comments and reformatting to follow same style than the rest of Fabulous.
> 
> I also added an `identicalElements` list so we don't need to remove an identical element (without a key) from reusableElements at the end.
> 
> > For current logic it looks like it is not needed, because we clear rest on create, and remove just does not take place because it is useless in such scenario. So, we make create and do not need remove because there is nothing to remove
> 
> Even though I agree with you that it doesn't change anything for the algorithm, I prefer to have create + remove instead of just create.
> It makes the intent clearer. We can't reuse the previous element, so we create a new one and delete the previous one.

delete is a  "fake" operation in such cases, because create already acts as create+ replace. If create was using insert instead .[i]<- x, then deletion would be needed

TimLariviere:
Yes absolutely. This is just needed for a "complete picture" in the tests.
`updateCollectionGeneric` completely ignores it.

TimLariviere:
@vshapenko Ok, finally got it working!

I had to fix 2 issues to make TicTacToe work correctly:

- When applied to `ObservableCollection`, the algorithm could potentially throw, even though it should have been fine.

That's because of the way Xamarin.Forms reacts to changes in `ObservableCollection`.
When we're moving an element from index X to Y (where X > Y), we set the index Y with the updated control and left X because it will be replaced by something else later.
Though Xamarin.Forms doesn't like having the same ref twice at different indexes, and it will try to delete the ref at X while we're trying to set at Y.
This resulted in Xamarin.Forms trying to update `ObservableCollection` while it was already being updated. Thus crashing.

I had to change the algorithm a bit to use `RemoveAt`/`Insert` instead of setting at a specific index.

- Updating attached properties (e.g. Grid.Column) was no longer working

Attached properties in Fabulous are tricky.
In the case of Grid, the attached properties are to be applied to the `Children` items but this property is defined in the inherited class `Layout<'T>`.
So the way CodeGen generated code required us to reprocess the "derived" attached properties (Grid => Layout<'T>) independently.
This was possible because the previous algorithm acted predictability by only using indexes.
It was not the case anymore.

So I changed the way attached properties were updated to make it possible to update them while executing `updateCollectionGeneric`.

vshapenko:
That rabbit hole is deeper than i thought

TimLariviere:
Well, for the attached properties, it was bound to happen I guess.
What was implemented always felt more like workaround than proper support.
It's slightly better now at least.

vshapenko:
Maybe would be worth to implement Replace method in observable collection in xf. This will give us one render pass instead of two in some cases

vshapenko:
@TimLariviere , what else do we need to merge this PR? If needed, would write a bunch of docs about keys

TimLariviere:
@vshapenko Yes, only the documentation remains.
It would be great if you write some docs about `key`, why is it needed, how to use it, what it will do concretely.
You can put that in https://fsprojects.github.io/Fabulous/Fabulous.XamarinForms/views-perf.html

vshapenko:
@TimLariviere , i have added documentation draft, may be some samples would be needed.

docs/Fabulous.XamarinForms/views-perf.md
```diff
@@ -60,16 +60,28 @@ The perf of incremental update to these is progressively less important as you g
...
+4. As final phase, we delete all controls belongs to elements in 'rest' and 'key' pools, which were not involved in update process.
 
 This means
 1. Incremental update costs minimally one transition of the whole list.
-2. Incremental update recycles visual elements at the start of the list and handles add/remove at end of list relatively efficiently
-3. Returning a new list that inserts an element at the beginning will recreate all elements down the way.
-
-Basically, incremental update is faster if items are being added/removed at the end, rather than the beginning of the list. 
+2. Incremental update recycles visual elements as much as possible if you use 'key' attribute.
+   Otherwise, there is no guarantee that you get same visual element next time.
```
TimLariviere, line 85: Maybe add a note about `ObservableCollection` here

```suggestion

NOTE: The list diffing will limit mutations to only Move, Remove, and Insert, even when more straightforward operations could be done.
This is to support the limitations imposed by how Xamarin.Forms reacts to changes in `System.Collections.ObjectModel.ObservableCollection<'T>`.

```

vshapenko:
@TimLariviere , updated documentation a little bit. Hope the purpose of key became more understandable

docs/Fabulous.XamarinForms/views-perf.md
```diff
@@ -47,6 +47,39 @@ You can also use
...
+     ```
+         
+  As we can see here, we have added a 'key' parameter to each Label.
+  Key should be unique across view elements in same collection, i.e.
+  Children of StackLayout and so on.
+  
+  2.What is a purpose of key? 
+   Purpose of key is following: 
+  * make determination of proper view element to replace easier(just by key,yes)
+  * most important,to keep a reference to previous XF control(so we would not create it, but reuse existing).
```
TimLariviere, line 75: I think we should explain the purpose of `key` from the user's point of view here.

When might the view not be performant?
=> Anything changing the ordering of controls from one update to the other: reordering, adding/removing elements at the start, etc.

Why is it a bad thing?
=> Fabulous will try to reuse existing controls as much as possible, in order by default. Changing the ordering might force Fabulous to recreate the controls which might impact the app.

How can I prevent that?
=> You can give intent to controls with the `key` property. Fabulous will try to use those intents to match controls having the same intent between updates, reusing controls more effectively.

For example:
```fs
View.StackLayout([
    if model.ShowFirstVideo then
        yield View.MediaElement(source = MediaPath "path/to/video.mp4")

    yield View.MediaElement(key = "other-video", source = MediaPath "path/to/other-video.mp4")
    yield View.Button(text = "Toggle first video", command = (fun () -> dispatch ToggleFirstVideo))
])
```

In this case, when `ShowFirstVideo` = false, the StackLayout will have 2 children, the 1st and 2nd video players.
When `ShowFirstVideo` = true, there will be only 1 child left, the 2nd video player.

Due to how Fabulous reuses views between updates, when switching from `ShowFirstVideo` true to false, Fabulous will remove the 2nd video player and reuse the 1st video player (the source will be the 2nd path), which might lose the current state of the 2nd player if it was playing.
That's because Fabulous has no way of knowing the real intent behind your code.

You can help it by specifying the `key` property which will let Fabulous know that the 2nd video player should still be there after the update.
Fabulous will only remove the 1st video player and keep the 2nd video player.

docs/Fabulous.XamarinForms/views-perf.md
```diff
@@ -47,6 +47,39 @@ You can also use
...
+  
+  In many scenarios this is essential thing, for example, you draw a control
+  that streams video, and its position changed due some conditions.Without 'key' set,
+  we get continuous recreating of control with a bunch of issues like video interrupt.
+  With key, we continue reuse  same control on each view cycle.
```
TimLariviere, line 80: We should say the opposite.

In the majority of scenarios, Fabulous will be efficient without doing anything specific.
In some advanced scenarios, like the one above, `key` will let you help Fabulous be even more efficient.

PR #726, opened 2020-04-30
Add support for Xamarin.Forms 4.6

TimLariviere:
Xamarin.Forms 4.6 has been released yesterday.
It contains 2 new controls: Expander and RadioButton as well as support for AppTheme (Light and Dark).
I added samples for those 3 in the AllControls sample.


Fabulous.XamarinForms/samples/AllControls/AllControls/Samples/UseCases/AppTheming.fs
```diff
@@ -0,0 +1,46 @@
...
+    
+    let view model _ =
+        match Device.RuntimePlatform with
+        | Device.iOS | Device.Android ->
+            appThemeView model
+
+        | _ -> 
+            View.ContentPage(
+                View.StackLayout [
+                    View.Label(text = "Your Platform does not support RadioButton")
```
rastreus, line 43:
```suggestion
                    View.Label(text = "Your Platform does not support AppTheming")
```
This looks like a simple copy-paste error from ```RadioButton.fs```.

TimLariviere: Thanks!

TimLariviere:
Not sure how AppTheme is supposed to work.
Tested it on both iOS and Android, and it doesn't work.

iOS never triggers `Application.Current.RequestedThemeChanged` when the phone theme changed.
Android recreates the activity hosting XF/Fabulous, thus resetting the state of the app to 0.

gaelian:
Apologies for replying to an old issue but it came up in my search. I'm currently trying to figure out how to get light/dark themes working with my full Elmish style Fabulous app (multi-page, somewhat following the model of the FabulousContacts example app). I've managed to get the themes sorted out, mainly by the [F#-coded styling approach](https://fsprojects.github.io/Fabulous/Fabulous.XamarinForms/view-a-styling.html), plus some native styling that I needed to do in the Android project (via values and values-night, etc). The F# code is simple, for example:

```f#
module Style =
    let isDarkTheme = Application.Current.RequestedTheme = OSAppTheme.Dark

    let TabBackgroundColor = if isDarkTheme then Color.FromHex("#2d2d30") else Color.FromHex("#f6f6f6")
    let TabUnselectedColor = if isDarkTheme then Color.FromHex("#d0d0d0") else Color.FromHex("#949594")
```

And these values are used within the various view functions to set colours on elements.

The themes are working, except that I can't apply them at runtime, e.g. if the theme changes automatically due to time of day or if the user manually changes the theme while the app is still running. The theme is only set when the app starts. I was hoping to use the `Application.Current.RequestedThemeChanged` event to dispatch a message (in conjunction with the approach discussed under [Triggering Commands from External Events](https://fsprojects.github.io/Fabulous/Fabulous.XamarinForms/update.html)) as I assumed that I would need to somehow trigger an update of the pages in the app for the F#-coded styling approach to reapply the different colours to the view elements. But as @TimLariviere mentions here, the `Application.Current.RequestedThemeChanged` event does not seem to fire.

If anyone could give any info or examples on how they have managed dynamically applying light/dark themes at runtime for Fabulous, that would be much appreciated.

Issue #728, opened 2020-05-03
Broken link to AllControls Samples section on Index of Docs


awright18:
Under Samples the "All Controls" link is broken. I believe it should link to App.fs at the same location. 

The sample [AllControls](https://github.com/fsprojects/Fabulous/blob/master/Fabulous.XamarinForms/samples/AllControls/AllControls/AllControls.fs) contains examples of instantiating most elements in Xamarin.Forms.Core.

PR #729, opened 2020-05-03
Fix issue #728 broken AllControls link

awright18:
Fixing the broken AllControls sample project link on the index page in the docs.


TimLariviere:
@awright18 Thank you for reporting and submitting a fix!
App.fs is not the most interesting file (it's mainly bootstrap for the control samples), so instead I changed it to redirect to the Samples folder with the actual code samples.

awright18:
Thanks Tim

Issue #733, opened 2020-05-05
0.55 - preview 1 : CollectionView broken


vshapenko:
Looks like items in CollectionView should be a kind of "special case" in new diff algo, because collection view looks completely broken.  It looks like virtualization breaks out hopes for better world...

TimLariviere:
@vshapenko Can you share a small sample so we can fix this, please?

vshapenko:
Ok, here is little one. Do not know how it is representative, but looks strange: 

```fsharp
 module App = 
    type Model = 
      { Items:int list }

    type Msg = 
        | Next 
    

   

    let initModel () = { Items = [1..100] }

    let init () = initModel () , []

    let update msg model =
        match msg with
        | Next -> model,Cmd.none

    let random = Random(DateTime.UtcNow.Ticks|>int)
    let view (model: Model) dispatch =
        let views =
            [
            
             for item in model.Items do
                 let kind = random.Next(0,100)
                 if(kind<20) then
                    yield View.Label(fontSize=FontSize 40.,text= sprintf "ITEM %i" item,backgroundColor =Color.Aqua)
                    
                 elif (kind<40) then
                     yield View.Frame(
                            cornerRadius = 24.,
                            backgroundColor = Color.Yellow,
                            content=View.Label(fontSize=FontSize 24.,text= sprintf "ITEM %i" item))
                 elif (kind<70) then
                     yield View.Label(fontSize=FontSize 24.,text= sprintf "Item %i" item,backgroundColor=Color.Red)
                 
                 else
                    yield View.Label(fontSize=FontSize (14.),text= sprintf "Item %i" item,backgroundColor=Color.Green) 
                 ]
        View.ContentPage(
          content=View.CollectionView(
            items = views
            
            ))
             
    let program = 
        Program.mkProgram init update view
```

vshapenko:
@TimLariviere , looks like this is a kind of view recyle behaviour, maybe we need to clear all previous attributes on update

TimLariviere:
@vshapenko I'm not seeing any difference between v0.54 and v0.55.0-preview1 using the sample you provided.
Am I missing something?

![versus](https://user-images.githubusercontent.com/6429007/81306761-2178cd00-9080-11ea-8cbd-14241aa90eaa.gif)

vshapenko:
@TimLariviere , hmmm. I will try to dig deeper. Problem is my app has rich ui inside collection view, i will think how to repeat it

[...]

PR #748, opened 2020-05-20
Fix an issue in BindableHelpers.onBindingContextChanged

TimLariviere:
Potential fix for #733 
Fixed an issue in `BindableHelpers.onBindingContextChanged` and also separated the collection diffing algorithm into 2 kinds: Children/Items.

The diffing algorithm for Items is less aggressive on reuse to avoid unwanted reordering animation, and will only reorder for same instance or same key, otherwise it only tries to reuse on the same index.

The diffing algorithm for Children will try to reuse all elements, even if the indexes are not matching.

### DIFF
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
@@ -82,9 +82,7 @@ module BindableHelpers =
                 newHolder.ViewElement.UpdateInherited(prevModelOpt, newHolder.ViewElement, bindableObject)
                 holderOpt <- ValueSome newHolder
                 prevModelOpt <- ValueSome newHolder.ViewElement
-            | _ ->
-                holderOpt <- ValueNone
-                prevModelOpt <- ValueNone
+            | _ -> ()
             
         onBindingContextChanged
         
--- a/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
+++ b/Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
@@ -82,7 +82,11 @@ module BindableHelpers =
                 newHolder.ViewElement.UpdateInherited(prevModelOpt, newHolder.ViewElement, bindableObject)
                 holderOpt <- ValueSome newHolder
                 prevModelOpt <- ValueSome newHolder.ViewElement
-            | _ -> ()
+            | _ ->
+                // Only remove the previous holder reference since we don't need it anymore
+                // prevModelOpt on the other hand needs to be kept, since Xamarin.Forms will first
+                // set BindingContext to null before setting a new item
+                holderOpt <- ValueNone
             
         onBindingContextChanged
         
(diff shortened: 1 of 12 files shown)


Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
```diff
@@ -82,9 +82,7 @@ module BindableHelpers =
                 newHolder.ViewElement.UpdateInherited(prevModelOpt, newHolder.ViewElement, bindableObject)
                 holderOpt <- ValueSome newHolder
                 prevModelOpt <- ValueSome newHolder.ViewElement
-            | _ ->
-                holderOpt <- ValueNone
-                prevModelOpt <- ValueNone
+            | _ -> ()
```
TimLariviere, line 85: @vshapenko This was the main culprit.
Whenever Xamarin.Forms tries to reuse a row, it first set the BindingContext of that row to `null` before setting it the new item.
And since we were resetting the `prevModelOpt` of the row on `null`, when the new item was set into `BindingContext`, Fabulous would think the row was empty and the diffing was incorrect.

vshapenko: Sounds brilliant. Never dived so deep.
Some time ago i tried to write an updated version of diff algo. The main idea was to  keep a collection of 

```fsharp
   type Element<'T> =
        | Direct of 'T
        | FromCell of int
 ```
where 'T is TTarget. So, main cycle just fills this list/array/whatever. Second pass maps to proper 'TTarget. and finally we compare elements of two 'TTarget by identical and swap if needed.  Here is the code: 

```fsharp
 type Element<'T> =
        | Direct of 'T
        | FromCell of int
        
    /// Incremental list maintenance: given a collection, and a previous version of that collection, perform
    /// a reduced number of clear/add/remove/insert operations
    let updateCollectionGeneric
           (prevCollOpt: 'T[] voption) 
           (collOpt: 'T[] voption) 
           (targetColl: IList<'TargetT>) 
           (create: 'T -> 'TargetT)
           (attach: 'T voption -> 'T -> 'TargetT -> unit) // adjust attached properties
           (canReuse : 'T -> 'T -> bool) // Used to check if reuse is possible
           (keyOf:'T -> string voption)
           (update: 'T -> 'T -> 'TargetT -> unit) // Incremental element-wise update, only if element reuse is allowed
        =
        let elems =
            match collOpt with
            | ValueSome v->Array.init<Element<'TargetT> option> v.Length (fun _->None)
            | ValueNone -> [||]
         
        let create index child =
            let targetChild = create child
            attach ValueNone child targetChild
            elems.[index]<-Direct targetChild |> Some
            
        let update index prevChild newChild =
            let targetChild = targetColl.[index]
            update prevChild newChild targetChild
            attach (ValueSome prevChild) newChild targetChild
            elems.[index]<-Direct targetChild |> Some
            
        let move prevIndex newIndex =
            elems.[newIndex]<-FromCell prevIndex |> Some
        
        updateCollectionGenericInternal
            prevCollOpt
            collOpt
            keyOf
            canReuse
            (fun () -> targetColl.Clear())
            create
            update
            move
            (fun index -> targetColl.RemoveAt(index))
            
        let newElems =
            [|
                for i in 0..elems.Length-1 do
                    match elems.[i] with
                    | Some (FromCell idx) ->
                        yield targetColl.[idx]
                    
                    |Some (Direct el) ->
                        yield el
                    
                    |None ->
                        yield targetColl.[i]
                    
            |]
        
        for i in 0..newElems.Length-1 do
            if(i>=targetColl.Count) then
                targetColl.Add newElems.[i]
            else 
                if(not (identical newElems.[i] targetColl.[i]))
                 then
                     targetColl.RemoveAt(i)
                     targetColl.Insert(i,newElems.[i])
```

Main idea is to simplify (and minimize) operations with target coll. Also in such implementation currState is not needed, as we can see, because we do not change collection before we have full list of updates

PR #766, opened 2020-06-26
Fix attached properties

TimLariviere:
Attached properties like `Grid.Column`, `Grid.ColumnSpan` had invalid default values because Fabulous was using the designated helper (`Grid.SetColumn(child, value)`) that has no "unset" equivalent.

This led to crashes in certain conditions.

So instead, Fabulous will now use directly the BindableProperty field (e.g. `Grid.ColumnProperty`) so it can clear the value if needed.


Happypig375:
Should `Fabulous.CodeGen` be renamed to something like `Fabulous.XamarinForms.CodeGen` instead? It should be under `Fabulous.XamarinForms`.

TimLariviere:
Fabulous.CodeGen generates by default code compatible with XAML-based frameworks (WPF, Uno, Avalonia, XF, ...) so it isn't necessarily depend on Xamarin.Forms.

All the XF specifics are handled by Fabulous.XamarinForms.Generator.

TimLariviere:
... and thanks. That let me realize I have wrote "Xamarin.Forms" is in the shared code. :)

Issue #738, opened 2020-05-10
Experiments: Stronger typing, Mixing Dynamic/Static views, Componentization, SwiftUI-like DSL 


TimLariviere:
DISCLAIMER: All the following points are experiments only and there's no guarantee they will be available anytime in the future.

Index:
- [Introduction](https://github.com/fsprojects/Fabulous/issues/738#issue-615417376)
- [Stronger typing & attribute-based diffing](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626348299)
- [Merging Fabulous and Fabulous.StaticViews](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626352462)
- [Running Fabulous as a component](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626354644)
- [Why should we change the Fabulous DSL?](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626355184)
- [Why not use Elmish.React / Feliz / Avalonia.FuncUI DSL instead?](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626389520)
- [SwiftUI-like DSL](https://github.com/fsprojects/Fabulous/issues/738#issuecomment-626395314)

In the last few days, I've been experimenting with a couple of ideas.

### Introduction

- **Stronger view typing (#341)**

Instead of having all `View.XXX()` return untyped `ViewElement`, it should return a typed ViewElement to make the compiler check for invalid uses (like putting a Button as an item of ListView -- ListView only accepts Cell-derivated controls).

- **Diffing per "attribute" instead of per "control"**

Currently in Fabulous.XamarinForms, we diff controls by calling the Update method attached to it (e.g. `View.Button()` will have a link to `ViewBuilders.UpdateButton()`).

This update method will check each one of the properties of Button (and will call UpdateVisualElement, etc. to update the inherited properties/events).
This is perfectly fine in 99% of the cases, except when dealing with overridden properties or  attached properties.

This becomes particularly tricking to always apply the correct behavior and it results in a lot of custom code.

Instead if we went with "attribute"-based diffing like FuncUI, overridden properties could simply be dealt with by storing the overridden attribute updater instead of the base one in an attributes array (e.g. ViewElement.Attributes).
Same with attached properties, there would be no need to be aware that `Grid.Row` is an attached property applicable only on the Grid inherited `Layout<T>.Children` items.

- **Merging Fabulous and Fabulous.StaticViews**

[...]

Issue #308, opened 2019-02-02
Button Text changes should not be animated


praeclarum:
When setting the view, nothing gets animated (correctly) except Button.Text gets animated (incorrectly). This creates several visual distractions including:

1. Appearance of an ellipsis if the new title is shorter
2. A quick fade out and fade back in (on iOS) that looks like a flicker
3. Text moving if the text length changes

This is especially bad if your UI has a lot of buttons because now everything is flickering.

For comparison, note how Label and Entry text is not animated.

praeclarum:
TBH, I'm worried this might be a bug with Forms. Need to find out what's triggering the animation.

TimLariviere:
It's not an issue in Fabulous itself. I reproduced it in both Fabulous and C# Xamarin.Forms.

Fabulous
```fsharp
// Copyright 2018-2019 Fabulous contributors. See LICENSE.md for license.
namespace TestPraeclarum

open System.Diagnostics
open Fabulous
open Fabulous.XamarinForms
open Xamarin.Forms

module App = 
    type Model = 
      { Toggled: bool }

    type Msg = 
        | Toggle

    let fullText = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aenean tempus vestibulum justo eu porttitor."
    let smallText = "Lorem"

    let init () = { Toggled=false }

    let update msg model =
        match msg with
        | Toggle -> { model with Toggled = not model.Toggled }

    let view (model: Model) dispatch =
        View.ContentPage(
          content = View.StackLayout(padding = 20.0, verticalOptions = LayoutOptions.Center,
            children = [ 
                View.Button("Toggle", command = (fun () -> dispatch Toggle))
                View.Button(if model.Toggled then fullText else smallText)
                View.Label(if model.Toggled then fullText else smallText)
                View.Entry(if model.Toggled then fullText else smallText)
            ]))

    let program = Program.mkSimple init update view

type App () as app = 
    inherit Application ()

    let runner = 
        App.program
        |> XamarinFormsProgram.run app
```

Xamarin.Forms
```xml
<?xml version="1.0" encoding="utf-8"?>
<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
             xmlns:d="http://xamarin.com/schemas/2014/forms/design"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
[...]

PR #775, opened 2020-07-10
Fix for v0.55 collections diffing

TimLariviere:
Reverted to a diffing that returns all changes necessary to move from state A to state B.

The main reason we did what we did before is that Xamarin.Forms needs we change ObservableCollection step by step, which was making the "old indices" really hard to calculate (on insert, all indices after are shifted by +1; on delete, shifted by -1).

So instead a new function is responsible to compensate for that, it's easier to understand what's going on.


vshapenko:
I think we can make the key attribute for collections mandatory. Something like items:(ViewElement*string) list. That will give some pain on update, but will let us have more proper collection diff.

Happypig375:
A `Map<string, ViewElement>` instead could be more efficient.

TimLariviere:
@vshapenko Yeah, maybe. Today, I think it wouldn't make sense API-wise though.
There would be 2 ways to declare keys and it would make no sense from the developer's point of view.

```fs
View.CollectionView([
    "Item1", View.Label(key = "Item2") // What's supposed to happen here?
])
```

Also I'm not a fan of having "dangling" values in the middle of controls.
I find it harder to understand. It feels loosely coupled in my opinion.
For example ListViewGrouped, without knowing the actual signature it's hard to understand what is the purpose of the different parts.
```fs
View.ListViewGrouped([
    "A", View.TextCell("A"), [
        View.TextCell("Alphabet")
    ]
])
```
As far as I know React doesn't require a key for collections, but strongly encourage it.
SwiftUI took a different approach by requiring the collection of data to have some kind of identifiable part, either by having the data type to implement a `Identifiable` protocol, or passing a lambda to the key as part of the `List` signature. 

```swift
List(landmarkData, id: \.id) { landmark in  //\.id here means fun data -> data.id
    LandmarkRow(landmark: landmark)
}
```

Which could be translated to
```fs
let items = [ { Id = 1; SomeData = "Hello" }; { Id = 2; SomeData = "World" } ]

View.CollectionView(
    itemsSource = items,
    itemId = (fun item -> item.Id),
    itemTemplate = (fun item ->
        View.Label(item.SomeData)
    )
)
```

Maybe it can be something we add to the new architecture as part of #738?

vshapenko:
I think yes. FYI, we have our own implementations of collection views. PM
me if you need details.

пн, 13 июл. 2020 г., 12:52 Timothé Larivière <notifications@github.com>:

TimLariviere:
I added a new constraint to prevent reuse in some conditions.
A previous element with a key that is not found in the new list won't be reused by a new element.

Even though nothing prevents us to reuse the abandoned keyed element, on some platforms, an animation might play on some properties when changed.
This is the case for iOS Button. When the text changes, it plays a fade-in/fade-out animation (#308).
Same with Android Button. When clicked, a ripple animation plays.

But due to Fabulous reusing those controls "aggressively" without the developer being able to say anything, it resulted in undesired animations playing.

Now, by default Fabulous will reuse an unkeyed control, but if the developer notices an undesired animation due to control reuse, it can disable it by setting a key.

vshapenko:
@TimLariviere , looks like we badly need your help

PR #780, opened 2020-07-30
Add extraAttributes support

Dolfik1:
At this moment we can add extra attributes via [WithAttribute](https://github.com/fsprojects/Fabulous/blob/master/src/Fabulous/ViewElement.fs#L183) function. This way is very slow because we duplicate element on each `WithAttribute` call. This PR adds `extraAttributes` key to all views:

```fsharp
View.Label(
  text = "Hello, world!",
  extraAttributes = [
    key "helloworld"
    row 0
    column 1
  ]
)
```
`extraAttributes` preallocates space in beginning of array and call `WithAttribute` function for each value in `extraAttributes`.


SergejDK:
I discussed this with @TimLariviere and this will be available in Fabulous v2 which he currently works on.

I will close this PR because this is planned in v2 with a new way for handling this.

PR #793, opened 2020-08-20
Add migration guide to v0.57 to contents.md

Happypig375:
So that it will be shown when the Views pages are collapsed.


TimLariviere (approved):
Ah thanks! I knew I was missing something 😅

PR #797, opened 2020-08-26
Version constraints in dotnet pack

### LINKED ISSUE #795: Fabulous nuget pacakge not listing an FSharp.Core dependency
I notice the FSHarp.Core dependency is not being listed on the Fabulous nuget package, I'm not sure why, I guess it just needs an entry in the nuspec

https://www.nuget.org/packages/Fabulous/0.57.0

TimLariviere:
Closes #795 

A bit of a hacky solution to force a `[lowerBound, upperBound)` version constraint in the nupkgs for `ProjectReference`.


TimLariviere:
Found a less hacky way to do the exact same thing: https://github.com/NuGet/Home/issues/5556#issuecomment-585482714

PR #798, opened 2020-08-27
fix pedantic typo in various AboutAssets.txt

cdituri:
"deployed with you package" => "deployed with your package"

Fabulous.XamarinForms/samples/FabulousWeather/Droid/Assets/AboutAssets.txt
Fabulous.XamarinForms/samples/LoginShape/Droid/Assets/AboutAssets.txt
Fabulous.XamarinForms/samples/ShapesDemo/Droid/Assets/AboutAssets.txt
Fabulous.XamarinForms/templates/content/blank/NewApp.Android/Assets/AboutAssets.txt

Signed-off-by: Chris Dituri <csdituri@gmail.com>

### DIFF
--- a/Fabulous.XamarinForms/samples/FabulousWeather/Droid/Assets/AboutAssets.txt
+++ b/Fabulous.XamarinForms/samples/FabulousWeather/Droid/Assets/AboutAssets.txt
@@ -1,7 +1,7 @@
 Any raw assets you want to be deployed with your application can be placed in
 this directory (and child directories) and given a Build Action of "AndroidAsset".
 
-These files will be deployed with you package and will be accessible using Android's
+These files will be deployed with your package and will be accessible using Android's
 AssetManager, like this:
 
 public class ReadAsset : Activity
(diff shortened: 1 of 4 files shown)


cdituri:
Another alternative would be to use git symlinks. That way there is one source of truth for `AboutAssets.txt`; and git symlinks are cross-platform across Windows, OS X, and Linux. The immediate downside is that all developers would need to ensure their git config has `core.symlinks = true`, which is not the git default.

```
src\Fabulous> git config -l | select-string ".*symlink.*"
core.symlinks=false
```

What's less clear to me is if using git symlinks would break any of Fabulous' internal plumbing or generation functionality. First time Fabulous user, and already loving it. Sending my thanks to the creators, maintainers, and contributors!

TimLariviere (approved):
Thanks!

>Another alternative would be to use git symlinks. (...) The immediate downside is that all developers would need to ensure their git config has core.symlinks = true, which is not the git default.

I'm not a fan of using symlinks because like you pointed out, all people cloning the repo would need to check their Git settings, which is not really obvious.

Also this file in particular is not really worth it. It's only there for information. It only needs to be shipped with the template, so we can delete it in the samples and only have one copy in the entire repo.

cdituri:
Makes sense @TimLariviere. :+1: on everything you said re: symlinks and shipping the single template, while removing the samples copies. Thanks again!

Issue #307, opened 2019-01-31
Consider using Debug.WriteLine instead of Console.WriteLine/printfn


TimLariviere:
Most output generated by Fabulous goes to the console through `Console.WriteLine` or `printfn`.
For platforms like WPF, Visual Studio doesn't show console outputs into the Output window.

`Program` is extensible to use a tracing function that can print elsewhere, but some outputs from Fabulous don't go through this function.
We need to provide an extension point where the developers can get all the messages from Fabulous.

`Debug.WriteLine` seems a good candidate.
It can be extended using Listeners like
```fsharp
System.Diagnostics.Debug.Listeners.Add(new ConsoleTraceListener()) |> ignore
System.Diagnostics.Debug.WriteLine("Hello")
```

edgarfgp:
I would like to contribute to this project. And this look like a good first one due to my knowledge of fsharp let me know if this still available .

TimLariviere:
@edgarfgp Yes. It's still available.
Feel free to contribute on this one!

The main issue is mostly with WPF.
When executing it, the output is not visible in the Debug window of Visual Studio.
But when executing it as a Console app (can be changed in the Properties of the WPF project), the output is visible in the opened Console.

I think it's because we are using `Console.WriteLine()` and not `Debug.WriteLine()`.

edgarfgp:
Update : I'm currently working in this issue

TimLariviere:
Great! :)
No, nothing particular has changed regarding this issue. It's basically the same code.

PR #809, opened 2020-10-27
Rework ViewElement to open new usages

TimLariviere:
As part of the allocations reduction process, I'm reworking the `ViewElement` class to become an interface `IViewElement`.
This will allow for different implementation such as `ComponentViewElement` (see https://github.com/TimLariviere/Fabulous/issues/25 for more information).

In this example, `ComponentViewElement` would drastically reduce the number of allocated ViewElement at each update.


TimLariviere:
Introduced `Registrar` and `DynamicViewElementHandler` types.
Those will avoid having to store `targetType`, `create`, `update` and `updateAttachedProperties` inside each `DynamicViewElement` instances, reducing the memory footprint of 1 `DynamicViewElement`.

So this should have a great impact on the memory allocation due to the sheer numbers of `DynamicViewElement` instantiated throughout the life of an application.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
```diff
@@ -15,96 +15,101 @@ open System.ComponentModel
...
     inherit INotifyPropertyChanged
-    abstract ViewElement : ViewElement
+    abstract ViewElement : IViewElement
+    abstract ProgramDefinition : ProgramDefinition
 
 [<AllowNullLiteral>]
-type ViewElementHolder(viewElement: ViewElement) =
-    let ev = new Event<_,_>()
+type ViewElementHolder(viewElement: IViewElement, programDefinition: ProgramDefinition) =
+    let ev = Event<_,_>()
```
dsyme, line 23: You might also want to populate this event object on-demand (mutable, null initially)

src/Fabulous/IViewElement.fs
```diff
@@ -0,0 +1,23 @@
...
+open Fabulous.Tracing
+
+type ProgramDefinition =
+    { CanReuseView: IViewElement -> IViewElement -> bool
+      Dispatch: obj -> unit
+      TraceLevel: TraceLevel
+      Trace: TraceLevel * string -> unit }
+
+and IViewElement =
+    abstract Create: ProgramDefinition -> obj
```
dsyme, line 14: I guess `///` comments on each of these would be good :-) Given the central nature of the interface :)

src/Fabulous/Runner.fs
```diff
@@ -0,0 +1,116 @@
...
+open Fabulous.Tracing
+open System
+
+type RunnerDispatch<'msg>()  =
+    let mutable dispatchImpl = (fun (_msg: 'msg) -> failwith "do not call dispatch during initialization" : unit)
+    member x.DispatchViaThunk = id (fun msg -> dispatchImpl msg)
+    member x.SetDispatchThunk v = dispatchImpl <- v
+
+/// Program type captures various aspects of program behavior
+type RunnerDefinition<'arg, 'msg, 'model, 'externalMsg> =
```
dsyme, line 13: TBH this should really have always been an interface I think.  I made it a record out of deference to the Elmish/Elm origins

dsyme:
@TimLariviere I love what you're doing here

TimLariviere:
Thanks!

src/Fabulous/IViewElement.fs
TimLariviere: Yes, I'm planning to put `///` comments on all public surface APIs.

Fabulous.XamarinForms/src/Fabulous.XamarinForms.Core/CustomControls.fs
TimLariviere: Thanks. Will take a look if that's possible. Not sure if Xamarin.Forms expects `PropertyChanged` right after the creation.

src/Fabulous/Runner.fs
TimLariviere: The record did not bother me because it's easy to extend with optional and/or custom behaviors to it through `Program.with...`.
An interface could do it too I guess, but might not be as simple to use.
I need to think about it.

TimLariviere:
This PR also rework the way we print some debug info to the console, but using a built-in tracing tooling.
This will help fix #307

PR #814, opened 2020-11-14
Update samples and templates

TimLariviere:
Update the samples to target Android 11, as well as the templates.

Things to do:
- [x] Re-enable Maps sample in AllControls
- [x] Ensure all samples fully work
- [x] Ensure templates work as expected on both macOS and Windows with the latest IDEs updates
- [x] Add System.Buffers to templates
- [x] Add System.Buffers to samples
- [x] Update documentation

- Ensure F# 5 can be used (Partial support)
    - Visual Studio (Windows) 16.8: OK
    - Visual Studio Mac 8.8 (Stable): Not supported at all
    - Visual Studio Mac 8.9 (Preview): Partial support (can build and run, but [no IDE support for new syntax](https://github.com/mono/mono/pull/20511#issuecomment-729184597) for the moment)


SergejDK:
@TimLariviere what is wrong with fabulousweather?

TimLariviere:
@SergejDK It was complaining that `Xamarin.Forms.targets` was imported 2 times when it was not. I got frustrated after fighting with it for 1 hour. 😅

I'll readd it after I'm done with the rest.

SergejDK:
@TimLariviere does this maybe have something to do with the other issue and android? maybe this is a bug, too from Xamarin.Android and the tooling?
Do you still have the older VS version installed and can check this ? Looks similar to me...

TimLariviere:
I don't think it's linked to the issue with VS.
I'm working on macOS with the latest versions of VS Mac and Rider.

I think it's due to an invalid state in the Fabulous solution that forces the IDEs to load the FabulousWeather project 2 times.

TimLariviere:
The hosted build agents are not yet updated to build .NET 5.0 apps...
Apparently, they will be updated next week.
https://developercommunity.visualstudio.com/content/problem/1255480/building-net-50-in-a-pipeline-doesnt-work.html

I will revert that change in the meantime.

PR #821, opened 2020-11-26
Issue Templates for Bugs / Feature Requests and Discussions

SergejDK:
@TimLariviere I just added some issue templates which could be helpful in future.

what do you think about it?


TimLariviere (approved):
Looks great!

Issue #727, opened 2020-05-01
ListView: disable animation when selectedItem changes


nghia-bui:
I have a ListView like this:

```
View.ListView (
                    selectionMode = ListViewSelectionMode.Single,
                    selectedItem = model.SelectedIndex,
                    itemSelected = (fun i -> i |> Option.iter (fun i -> dispatch <| UpdateSelectedIndex i)),
                    items = [ for name in model.Names do yield itemView name ]
)
```

Problem is that the items (with new content) are shown immediately in the UI, but the selected item is highlighted a little bit late due to the animation.

How to disable this animation?

SergejDK:
@nghia-bui just to be clear - you don't want to be able to highlight any row in the list?

PR #822, opened 2020-12-02
Added custom renderer doc

SergejDK:
Relates to #727


docs/Fabulous.XamarinForms/custom-renderer.md
```diff
@@ -0,0 +1,112 @@
+{% include_relative _header.md %}
+
+{% include_relative contents-views.md %}
+
+## Views: Custom Renderer / ViewElements
+
+Xamarin.Forms user interfaces are rendered using the native controls of the target platform, allowing Xamarin.Forms applications to retain the appropriate look and feel for each platform. Developers can implement their own custom Renderer classes to customize the appearance and/or behavior of a control.
+
+### Customized ViewElements
```
TimLariviere, line 9: I don't think the goal of those "customized ViewElements" is really clear. The TestLabel example is a technical one rather than a real-world use case.
Maybe it would be clearer to rephrase like:
>If an existing control does not provide the feature you want, you can extend it through a custom control.
>To do that, you can do this and that...

I just remembered a better example from FabulousContacts: https://github.com/TimLariviere/FabulousContacts/blob/master/FabulousContacts/Controls/UnderlinedLabel.fs

I created this UnderlinedLabel control because Xamarin.Forms does not provide underline option on Label while the native platforms support it (with the appropriate CustomRenderer)

TimLariviere: Also having a custom control without custom renderers is not really usual. Custom controls usually need custom renderers to apply the new features (like the underlining).
Though you can have just custom renderers without having to change the existing XF control.

TimLariviere: Don't hesitate to just redirect to the Xamarin.Forms documentation rather than explaining it yourself for all the Xamarin.Forms specific bits.
In addition to that, you can also point to one of our sample files (even FabulousContacts) for an example of "How it's done in F#".

docs/Fabulous.XamarinForms/custom-renderer.md
```diff
@@ -0,0 +1,112 @@
...
+        base.OnElementPropertyChanged(sender, e)
+
+        match e.PropertyName with
+        | "SelectedItem" ->
+            match this.Element, this.Control with
+            | null, null -> ()
+            | _ ->
+                this.Control.VisibleCells
+                |> Array.iter (fun v -> v.SelectionStyle <- UITableViewCellSelectionStyle.None ) // disable all selectionpossibilities
+        | _ -> ()
```
TimLariviere, line 70: Personally, I would recommend to rewrite it like:
```fs
if e.PropertyName = "SelectedItem" then
    for cell in this.Controls.VisibleCells do
        cell.SelectionStyle <- UITableViewCellSelectionStyle.None
```

Element and Control won't be null here, and functional style can sometimes be harder to read than "dumb" statements :)

TimLariviere (commented):
Thanks for taking time to write this documentation. It's indeed something we're missing. :)
I made some comments to make this a little bit clearer for newcomers.

TimLariviere (approved):

PR #865, opened 2021-04-10
Fix support for nested types

Dolfik1:


Dolfik1:
Hmm. It looks like build problem is not related to PR.

TimLariviere (approved):

PR #868, opened 2021-04-18
Docs

reigam:
My attempt to update the user docs.

My intentions:
* Make it more accessible for beginners (like me) by 
* Keep all prior content (just rearranged some things)
* Structure content more like the Xamarin.Forms documentations does
   * https://docs.microsoft.com/en-us/xamarin/xamarin-forms/user-interface/
   * https://github.com/reigam/Fabulous/blob/docs/docs/Fabulous.XamarinForms/images/guide-organisation.jpg
* Show which pages / topics are up to date 
  * e.g.  `topic last updated: v 0.61.0 - 31.03.2021 - 11:15am`


dsyme:
This looks great!

TimLariviere:
Thanks for the pull request!
I've been trying to make it run locally, but I found a few issues:
- Some md files include a reference to the deleted `contents-views.md`
- Apparently, you can't use `../` to include files in parent folders (https://jekyllrb.com/docs/includes/#including-files-relative-to-another-file)

reigam:
Ok, I checked and (hopefully) corrected all broken .md links.
I also moved all .md pages to the same folder, so that should be fixed. 
Unfortunately I noticed your P.S. too late. I will comment again after I checked the docs locally myself. 

Thanks for your feedback.

reigam:
Ok, I have made some some corrections (changing links from .md to .html mainly), and now everything runs fine locally on my machine.

TimLariviere (approved):
I think there might be still some dead links here and there, but we will fix them later.

PR #872, opened 2021-05-05
Fixed broken links because of mismatched upper and lower cases

reigam:


reigam:
Looks like the broken links resulted from mismatching upper and lower cases. Should be fixed now. 
Couldn't find a way to make Git recognize changes in the image names when there were only changes from lower to upper case, so I had to rename the folder (views to view).

TimLariviere (approved):

PR #880, opened 2021-06-15
Fix VisualStudio v16 Error

SergejDK:
We got an error with updating to VS 16 / VsMac 8.10 and it failed due to `MSBuild.StructuredLogger`.

This is just a workaround for now -> https://github.com/fsharp/FAKE/issues/2595


SergejDK:
@TimLariviere do we need to include this into the v1.0 branch?

TimLariviere:
Yes, it would be good to apply that commit to v1.0 too

PR #882, opened 2021-06-17
Fix Template issue with WPF/UWP

### LINKED ISSUE #881: [Bug]  New App from template UWP/WPF FileNotFoundException
### Steps to Reproduce

1. Create a new app -->  dotnet new fabulous-xf-app -n SqueakyApp --WPF --UWP
2. Open .sln in VS 2019
3. Run WPF or UWP on your local computer
4. `System.IO.FileNotFoundException: "Could not load file or assembly 'Fabulous, Version=1.1.0.0, Culture=neutral, PublicKeyToken=null'. `

### Expected Behavior
Runs without an error

### Actual Behavior
`System.IO.FileNotFoundException: "Could not load file or assembly 'Fabulous, Version=1.1.0.0, Culture=neutral, PublicKeyToken=null'. `

### Basic Information

- Version with issue: Fabulous.XamarinForms 1.1.0 & Fabulous 1.0.0
- Last known good version: no information
- Platform Target Frameworks: 
  - UWP: 1809
  - WPF: 

### Environment

```
Microsoft Visual Studio Community 2019
Version 16.10.2
VisualStudio.16.Release/16.10.2+31410.357
Microsoft .NET Framework
Version 4.8.04084

Installierte Version: Community
```

### Build Logs

`System.IO.FileNotFoundException`  --> `Could not load file or assembly 'Fabulous, Version=1.1.0.0, Culture=neutral, PublicKeyToken=null'`

### Workaround

Adding the Fabulous Nuget package will fix the issue but we have to include it into the template files.

SergejDK:
@TimLariviere I think the problem with #881 was that the references for fabulous were missing. 

I included this into both template proj files.

closes #881

### DIFF
--- a/Fabulous.XamarinForms/templates/content/blank/NewApp.UWP/NewApp.UWP.csproj
+++ b/Fabulous.XamarinForms/templates/content/blank/NewApp.UWP/NewApp.UWP.csproj
@@ -151,6 +151,18 @@
   <ItemGroup>
     <PackageReference Include="FSharp.Core" Version="FSharpCorePkgVersion" />
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform" Version="MicrosoftNETCoreUniversalWindowsPlatformPkgVersion" />
+    <Reference Include="Fabulous">
+      <HintPath>..\packages\Fabulous.FabulousPkgsVersion\lib\netstandard2.0\Fabulous.dll</HintPath>
+    </Reference>
+    <Reference Include="Fabulous.LiveUpdate">
+      <HintPath>..\packages\Fabulous.LiveUpdate.FabulousPkgsVersion\lib\netstandard2.0\Fabulous.LiveUpdate.dll</HintPath>
+    </Reference>
+    <Reference Include="Fabulous.XamarinForms">
+      <HintPath>..\packages\Fabulous.XamarinForms.FabulousPkgsVersion\lib\netstandard2.0\Fabulous.XamarinForms.dll</HintPath>
+    </Reference>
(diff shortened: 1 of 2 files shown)


TimLariviere:
Weird it doesn't work anymore.

UWP and WPF projects use the SDK-Style format for csproj/fsproj so this implies several things:
- We should use `PackageReference` instead of `Reference` for NuGet packages.
- If a project references another project via `ProjectReference`, it should automatically reference all the `PackageReference`s from that other project.

And this is the case here: `NewApp.fsproj` imports the various Fabulous packages; `NewApp.UWP`/`NewApp.WPF` reference the project `NewApp`.

SergejDK:
So it seems like this is not the right way to go for this problem.

Found some other interesting parts which could help but I have to test them when I am back at my windows computer.

Is there an easy way to create uwp/wpf project from the templates folder?

TimLariviere:
> Is there an easy way to create uwp/wpf project from the templates folder?

I think now with .NET 5 there is a `dotnet new` template for WPF at least.

SergejDK:
@TimLariviere yes `dotnet new -i` can install a template from a path. I'm currently looking into this issue and trying different things. The packagereference will work but it is kinda dirty.

SergejDK:
@TimLariviere 
Found sth. interesting - `Fabulous.XamarinForms` version 1.1.0 references `Fabulous` 1.0. If I include Fabulous into the shared project so it will update to 1.1.0 it will work again. Even after removing Fabulous after that it still works on WPF/UWP. 

Do you have an idea where the reference is wrong then maybe I can fix that.

PR #900, opened 2021-11-24
Fix build for linux in azure-pipeline

SergejDK:
@TimLariviere 

I already had a branch where I tested a lot but I thought that it would be better to have a clean branch with the solution. 

Could you check this please?


TimLariviere (approved):
Awesome work, thank you! 👍

PR #926, opened 2022-05-12
Introduced the notion of small scalars (can fit into 8 bytes) and converted all XF attributes that made sense

twop:
Converted small scalars to use 8 bytes encoding when possible

I'm not 100% sure what is the best API is but I think this is a reasonable place to merge.

specifically I'm skeptical of names like these

```fsharp
Attributes.defineBool name update
Attributes.defineBindableBool attribute
```

But I haven't found a better design yet.

Converted all XF attributes to work with small scalars. I'm super happy that I found out a way to generalize working with enums

```fsharp
Attributes.defineBindableEnum<MyEnum> bindableProp
```

We have a gazzilion of those ^_^

I couldn't run sample apps (yet) but I believe everything should just work (famous last words)

### DIFF
--- a/src/Fabulous.XamarinForms/Attributes.fs
+++ b/src/Fabulous.XamarinForms/Attributes.fs
@@ -164,7 +164,7 @@ module Attributes =
     /// that is, it allocates 2 bytes for each of channel of RGBA.
     /// Technically it might loose precision of (0.0 .. 1.0) float range used in XF.Color.
     /// If you want to avoid any potential loss of accuracy you can use "defineBindable" instead.
-    /// It is 100% accurate but allocates XF.Color values on the heap (thus slower) 
+    /// It is 100% accurate but allocates XF.Color values on the heap (thus slower)
     let inline defineBindableColor (bindableProperty: BindableProperty) =
         defineSmallBindable bindableProperty SmallScalars.Color.decode
 
--- a/src/Fabulous.XamarinForms/Attributes.fs
+++ b/src/Fabulous.XamarinForms/Attributes.fs
@@ -1,5 +1,6 @@
 namespace Fabulous.XamarinForms
 
+open System.Runtime.CompilerServices
 open Fabulous
 open Fabulous.ScalarAttributeDefinitions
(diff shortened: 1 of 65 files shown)


TimLariviere:
This is looking very good 👍
Love the way you use overload to remove the explicit encode function.

src/Fabulous.XamarinForms/Attributes.fs
```diff
@@ -24,6 +25,59 @@ type ValueEventData<'data, 'eventArgs> =
...
     let create (value: 'data) (event: 'eventArgs -> obj) = { Value = value; Event = event }
 
+/// Xamarin Forms specific attributes that can be encoded as 8 bytes
+module SmallScalars =
+    /// In reality XF takes 7 values of 4 bytes each.
+    /// Which is larger than 8 bytes an unusual to represent a color
+    /// So we are just going to bluntly convert it into 8 bytes in rgba form
+    /// note that we allocate 2 bytes for each channel in RGBA format to improve precision
+    /// That being said, it is a lossy conversion
+    module Color =
```
twop, line 35: @TimLariviere this one I'm not sure about, I think it is worth it but the color value might not be exactly the same, that said it is going to be less than 0.001% accuracy loss, thus I think it is fine

TimLariviere: I think this is fine too. Most apps are usually "standard" colors anyway, where the accuracy loss won't be an issue.
The potential perf boost will surely be way more beneficial.

If some people really really need that complete accuracy, it will still be easy enough to declare another attribute using the obj casting.

TimLariviere: Also would be good to point out that accuracy loss in the XML comment of the color modifiers

twop: Added a comment

TimLariviere (approved):
Awesome! I'm merging it to be able to release a new preview.
If the preview is good, I will release a stable 2.0 🥳

PR #949, opened 2022-06-08
Make sure navigation events are not triggered by Fabulous

### LINKED ISSUE #947: NavigationPage crash
## Repro project
https://github.com/reigam/Fab2Demo

## Repro branch in Fabulous
https://github.com/fsprojects/Fabulous/tree/wip-bug-nav-page-crash-repro
Note that it has a bunch of printfn statement because I couldn't attach a debugger

## Steps
1. Start the app (either iOS or android)
2. Click "Go to page 2"
3. Click "Close all"
4. Click "Go to page 2" again on the fist page
5. Clcik "Close all" again on the second page
6. crash incoming!

## My thoughts
It seems that there is a logical disconnect somewhere in nav page wrapper for Fabulous. 

When there is an event "Popped" the update function clears up the navigation stack completely, and there are no pages in XF after "Clear All". Maybe that is a logical error that we should throw if there are no pages returned?

But also the "Popped" event can be triggered either from "Back Button", e.g. from user interaction in XF internals or via change in the stack property, thus it seems there is no definitive source of truth in here :(

TimLariviere:
Fixes #947 

This PR does a couple things:
- Mark NavigationPage's `onPopped`, `onPushed` and `onPoppedToRoot` as obsolete
- Introduce NavigationPage `onBackNavigated`
- Allow Fabulous to catch up with XF NavigationPage in case NavPage already popped a page Fabulous was not aware of

The main issue with #947 was all NavigationPage events were both triggered by user actions and programmatic changes made by Fabulous. There was no way to differentiate who was triggering the events to be able to react only to the user's actions.

To prevent this, I created `CustomNavigationPage` which has an internal counter for the number of times Fabulous pops pages. This counter is used to suppress the `Popped` events triggered programmatically by Fabulous.
If the pop has been done by the user, then `BackNavigated` is triggered instead.

### DIFF
--- a/samples/Xamarin.Forms/FabulousContacts/FabulousContacts/App.fs
+++ b/samples/Xamarin.Forms/FabulousContacts/FabulousContacts/App.fs
@@ -18,19 +18,14 @@ module App =
         | UpdateWhenContactAdded of Contact
         | UpdateWhenContactUpdated of Contact
         | UpdateWhenContactDeleted of Contact
-        | NavigationPopped
+        | BackNavigated
 
     type Model =
         { DbPath: string
           MainPageModel: MainPage.Model
           DetailPageModel: DetailPage.Model option
           EditPageModel: EditPage.Model option
-          AboutPageModel: unit option
-          /// This field counts the number of pages we are popping at once
-          /// This is done to be able to differentiate NavigationPage.onPopped triggered by us
-          /// or by users pressing back button.
-          /// Xamarin.Forms doesn't tell us the difference
-          PoppingCount: int }
+          AboutPageModel: unit option }
 
     let init dbPath =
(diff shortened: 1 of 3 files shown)


TimLariviere:
I tried it against the repro branch and it was working perfectly well once I replaced `onPopped` with `onBackNavigated`

edgarfgp (approved):

twop (approved):
It seems this is the best that we can do here.

I wonder if a simple counter is going to be enough and there is no chance for a race conditions.

E.g. while we are popping a page programmatically the user pressed back (or some other obscure scenario).

I have a feeling that this solution would work for 99.99% of time but there is a small chance of weird stuff happening.

Again, I don't have a better idea/suggestion. 

Awesome work!!

PR #955, opened 2022-06-17
View API Reference + Translate Extensions

edgarfgp:
This PR adds View API reference doc and 2 new translation extesnions


src/Fabulous.XamarinForms/Views/_View.fs
```diff
@@ -62,6 +62,26 @@ module XFView =
                     view.TranslateTo(data.X, data.Y, data.AnimationDuration, data.Easing)
                     |> ignore)
 
+    let TranslateXTo =
```
TimLariviere, line 65: I just noticed Translate/Scale/Rotate are actually from VisualElement instead of View. Could you move all those to VisualElement.fs, please?
https://docs.microsoft.com/en-us/dotnet/api/xamarin.forms.visualelement?view=xamarin-forms

src/Fabulous.XamarinForms/Views/_View.fs
```diff
@@ -62,6 +62,26 @@ module XFView =
...
+        Attributes.defineSimpleScalarWithEquality<float>
+            "View_TranslateXTo"
+            (fun _ newValueOpt node ->
+                let view = node.Target :?> View
+
+                match newValueOpt with
+                | ValueNone -> view.TranslationX <- 0.
```
TimLariviere, line 72: Unlike ScaleX/ScaleY/RotateX/RotateY, TranslationX/TranslationY don't have corresponding methods with animation.
And since they are BindableProperties, we need to call `view.SetValue(VisualElement.TranslationXProperty, value)` / `view.ClearValue(VisualElement.TranslationXProperty)` instead of direct assignments

TimLariviere (changes requested):

src/Fabulous.XamarinForms/Views/_View.fs
TimLariviere: I guess you can replace everything with `Attributes.defineBindableFloat`

src/Fabulous.XamarinForms/Views/_View.fs
edgarfgp: Yeah you are right. We were duplicating some of the extensions :) . Moved to VisualElement and delete the duplicate ones and right API docs for it

src/Fabulous.XamarinForms/Views/_View.fs
edgarfgp: Done :)

TimLariviere (approved):
Thanks 👍

PR #956, opened 2022-06-18
Don't eat exceptions thrown in MVU functions

TimLariviere:
This PR introduces a `LogException` function in the Logger available inside the Program.
It also rethrows exceptions after calling LogException.

We use a Mailbox to process new messages and it makes it hard to rethrow an exception with the proper stacktrace.
I noticed Elmish moved away from MailboxProcessor, maybe we need something similar.

### DIFF
--- a/src/Fabulous.XamarinForms/Program.fs
+++ b/src/Fabulous.XamarinForms/Program.fs
@@ -89,11 +89,10 @@ module ViewHelpers =
 
             Trace.WriteLine(message, traceLevel)
 
-        let logException (ex: exn) = Trace.WriteLine(ex.ToString(), "Error")
-
         { Log = log
-          LogException = logException
           MinLogLevel = LogLevel.Error }
+
+    let defaultOnException _exn = false
 
 module Program =
     let inline private define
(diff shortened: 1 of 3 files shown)


edgarfgp (approved):

edgarfgp:
Is there any benefit apart from rethrow using a non MailBoxProcessor approach ? 

Will the possible change still allow us to add support to run multiple programs in the future ?

TimLariviere:
>Is there any benefit apart from rethrow using a non MailBoxProcessor approach ?

Not sure. Removing Mailbox would also let us avoid switching threads all the time.
We would be able to run by default on the main thread and let Cmd decide to move to another thread or not.
eg. `Cmd.MainThread.ofMsg` (UI thread) vs `Cmd.ofMsg` (another thread)

I feel this would be a cleaner approach compared to today.

>Will the possible change still allow us to add support to run multiple programs in the future ?

Yes. In v1, we had a singleton dispatch mainly for LiveUpdate.
So here we would have an instance function that dispatches only to the associated runner.
You can still have multiple runners in a same app.

TimLariviere:
I replaced the mailbox with a concurrent queue and a while loop.
Seems to be working fine, but I still need to dogfood it this change before merging it

TimLariviere:
This PR is quite an important change with potential breaking behavior. I will merge it to v2.1-preview branch

PR #958, opened 2022-06-20
Fix an issue where CarouselView could crash because of ViewRef

TimLariviere:


src/Fabulous.XamarinForms/Views/Collections/CarouselView.fs
```diff
@@ -34,12 +34,18 @@ module CarouselView =
...
-                            EventHandler<IndicatorView>
-                                (fun _ indicatorView -> (node.Target :?> CarouselView).IndicatorView <- indicatorView)
+                            Handler<IndicatorView>
+                                (fun viewRef indicatorView ->
+                                    let carouselView = node.Target :?> CarouselView
+                                    if carouselView <> null then
+                                        carouselView.IndicatorView <- indicatorView
+                                    else
+                                        // The target has been destroyed, clean up the handler
+                                        (viewRef :?> ViewRef<IndicatorView>).ClearListeners())
```
TimLariviere: Might not be the best idea 🤔 
We only want to clean up the current handler, not all handlers of ViewRef because the IndicatorView might still be alive

edgarfgp: Would it make sense IsAlive concept in the ViewRef and then use that to clean up only the one we want ?

TimLariviere: The issue was with the ViewNode.Target here, not ViewRef.
Unfortunately, ViewNode.Target uses WeakRef but it doesn't warn us when the target has been disposed.
So we need to check for null and then remove the old event handler.

PR #961, opened 2022-06-22
Fix GetViewNode call to use passed parameter

tylerhartwig:
I encountered a bug while attempting to create a view node tree using another kind of object in the tree (other than the BindableObject base).

The bug is that this method forces view node collections to use the root tree's GetViewNode. Instead it should use the passed parameter as the other branches of the function do.

I tried to create a unit test to show the bug, however I could not get the view collection to trigger the condition. If this is still wanted I can commit those files and would gladly accept guidance on triggering the affected code from the Sketch tests.

### DIFF
--- a/src/Fabulous/Attributes.fs
+++ b/src/Fabulous/Attributes.fs
@@ -234,8 +234,8 @@ module Attributes =
                     Dispatcher.dispatchEventForAllChildren itemNode widget Lifecycle.Mounted
 
                 | WidgetCollectionItemChange.Update (index, widgetDiff) ->
-                    let childNode =
-                        node.TreeContext.GetViewNode(box targetColl.[index])
+                    let childNode = getViewNode (box targetColl.[index])
+//                        node.TreeContext.GetViewNode(box targetColl.[index])
 
                     childNode.ApplyDiff(&widgetDiff)
 


TimLariviere:
I would approach this problem differently.

Initially I didn't wanted to have this `getViewNode` parameter, only use the root TreeContext to do that.
Passing a hardcoded function is less flexible (can't override it) and can have inconsistent behavior (the children will still use the root TreeContext)

I think I forgot to remove this parameter when I put GetViewNode in the TreeContext.

What I think would be better to do is to override the GetViewNode at the root level and add a condition:

```fs
if (yourControl matches some conditions) then
    yourCustomGetViewNode(yourControl)
else
    // Call the default GetViewNode
    Fabulous.XamarinForms.ViewNode.Get(yourControl)
```

tylerhartwig:
This makes sense -- would that also imply the other branches in this function should be adjusted to match? @twop not sure if there was other motivation for these changes as well, in terms of performance or anything.

TimLariviere:
> would that also imply the other branches in this function should be adjusted to match?

Yes. You can remove the `getViewNode` parameter of the define function and fix the errors by replacing with the TreeContext.

tylerhartwig:
Just saw this was closed, I'll update eventually with your recommendation @TimLariviere life has been busy lately

PR #965, opened 2022-06-27
Improve use of FlexLayout by moving wrap to ctor

TimLariviere:


edgarfgp (approved):

src/Fabulous.XamarinForms/Views/Layouts/FlexLayout.fs
```diff
@@ -47,8 +49,17 @@ module FlexLayout =
...
+        static member inline FlexLayout<'msg>(?wrap: FlexWrap) =
+            let wrap =
+                match wrap with
+                | None -> FlexWrap.NoWrap
+                | Some v -> v
+
+            CollectionBuilder<'msg, IFlexLayout, IView>(
+                FlexLayout.WidgetKey,
+                LayoutOfView.Children,
+                FlexLayout.Wrap.WithValue(wrap)
```
twop, line 61: It might be better not to add an attribute  if there is no wrap argument was specified

PR #969, opened 2022-06-29
call base on overriden OnPropertyChanged/OnPropertyChanging

stroborobo:
Hey guys,

noticed that neither selected index nor title were updating while using the Picker control. Turns out the CustomPicker (and others) didn't tell the base class about the changed properties.

I'm really enjoying this new API, thank you so much for making Fabulous! <3


TimLariviere (approved):
Nice catch. Thanks a lot 👍

PR #971, opened 2022-07-04
add custom application to redirect lifecycle calls to event handler

stroborobo:
Hey Fabulous team,

added a custom application class so we can have an event handler for the application lifecycle methods. It's clearly inspired by the custom picker and others, not sure if this is the intended way, appears to be working for me. Feedback very welcome if I should do it differently.

Also, are the docs generated or should I put it in there manually?


stroborobo:
I'm periodically fetching updated data while the app is open, so I'm both fetching updates and starting a subscription on resume and cancel it's cancellationtoken on sleep.

TimLariviere (approved):
👍

PR #973, opened 2022-07-11
Application lifecycle: update builder extensions to use msg directly

stroborobo:
With the change to use EventNoArgs, the event handler expects the scalar attribute to contain the msg, but they were still initialized with a function unit -> msg, so the dispatch unboxing crashed.


TimLariviere (approved):
Oh yes, sorry. Forgot to change that as well 🤕

stroborobo:
No worries, I could've seen that there's an attribute definition function for this, next time I'll make sure to look beyond the xamarinforms namespace 😅

Thank you :)

Issue #927, opened 2022-05-16
[v2] Tracking missing documentation


TimLariviere:

edgarfgp:
@TimLariviere should we add a entry for widgets styling ?

h0lg:
[Recent updates to Fabulous.Avalonia](https://github.com/fabulous-dev/Fabulous.Avalonia/pull/206/files#diff-96cc63366bf7cc2c4a28682e1162d6b7df145d8142981c054c8f219e9131b813) triggered me to look into the `Program` module and try to understand the difference between `stateful`,  `statefulWithCmd` and `statefulWithCmdMsg`, which includes the concept of `CmdMsg` vs. `Cmd<'msg>` - also increasingly important to understand the Gallery samples in Fabulous Avalonia. Please add some documentation explaining these concepts, which I can't seem to find explained elsewhere.

dsyme:
This v2-era tracker is superseded by the current Fabulous 10 documentation plan in #1122. The new tracker covers maintained MAUI/Avalonia backends, current program/command concepts, testing, deployment, architecture, and migration guidance.

PR #978, opened 2022-07-20
Add styling documentation

TimLariviere:
Part of #927


docs/content/en/v2/development/styling.md
```diff
@@ -0,0 +1,375 @@
...
+```fs
+[<Extension>]
+static member inline SomeExtensionMethod(this: TypeForWhichWeWantToAddAnExtension) =
+```
+
+**Now, the most important part to understand:** when using a widget, Fabulous will actually give you an instance of `WidgetBuilder<'msg, 'marker>` where `'marker` will be a type representing the widget.  
+For example, `Label()` will give you `WidgetBuilder<'msg, ILabel>`,
+
+You might have noticed that in the code above, we are using `#ILabel` instead of `ILabel`.  
+The `#` (pound) symbol lets F# know that we actually want the shorthand modifier to be available to all widgets deriving from `ILabel`, not just `Label`.
```
edgarfgp, line 154: Should we add a link the the Flexible types feature ? https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/flexible-types

edgarfgp (commented):
Should we also mention that there is the .resources modifier that will apply global styling ?

TimLariviere:
Hmm yes, you're right. I should add the "legacy" way to do styling via `.resources()`, `.style()`

docs/content/en/v2/development/styling.md
TimLariviere: Good idea! Didn't even know that it was called a flexible type 😅

PR #981, opened 2022-08-29
Implemented taking only the latest attribute for a property

twop:
## Context
In WidgetBuilders we can easily have duplicate attributes
like Label("aha!").color("red").color("blue")
Diffing algorithm already uses stable sort for diffing attributes
Thus we need to leverage that fact

In particular we might have a bug in this context
prev:  `Label("aha!").color("red").color("blue")`
next:  `Label("aha!").color("red")`

After the diffing the resulting color should be "red", but previously
it was interpreted as "removed color property" instead

## Solution
I implemented `SkipRepeatingScalars.skip` functions that skips `color("red") from `color("red").color("blue")`
Effectively taking only the latest attribute for a particular property.

I used in `ScalarChangesEnumerator` but only for actual diff. It Is possible that it should be used for "AllRemoved" and "AllAdded" but it is unlikely to cause any problems (possible a field will be set twice for example)

### DIFF
--- a/src/Fabulous/WidgetDiff.fs
+++ b/src/Fabulous/WidgetDiff.fs
@@ -23,6 +23,41 @@ module EnumerationMode =
         | ValueNone, ValueSome next -> EnumerationMode.AllAddedOrRemoved(next, true)
         | ValueSome prev, ValueSome next -> EnumerationMode.ActualDiff(prev, next)
 
+module private SkipRepeatingScalars =
+    /// Context:
+    /// In WidgetBuilders we can easily have duplicate attributes
+    /// like Label("aha!").color("red").color("blue")
+    /// Diffing algorithm already uses stable sort for diffing attributes
+    /// Thus we need to leverage that fact
+    ///
+    /// In particular we might have a bug in this context
+    /// prev:  Label("aha!").color("red").color("blue")
+    /// next:  Label("aha!").color("red")
+    ///
+    /// After the diffing the resulting color should be "red", but previously
+    /// it was interpreted as "removed color property" instead
+    ///
+    /// This functions skips "color("red")" from "color("red").color("blue")"
+    /// Effectively taking only the latest attribute for a particular property
+    ///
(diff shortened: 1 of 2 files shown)


src/Fabulous/WidgetDiff.fs
```diff
@@ -23,6 +23,42 @@ module EnumerationMode =
...
+        let length = scalars.Length
+        // either the last element or out of bounds
+        if pos >= length - 1 then
+            pos
+        else
+            // that means that there is at least one more element ahead
+            let key = scalars.[pos].Key
+            let mutable resultingIndex = pos
+
+            while (length - 1 > resultingIndex)
```
TimLariviere, line 56: Not sure if it makes any difference, but maybe a simple `if scalars.[pos + 1].Key = key then` would be better than a while condition on the hot path?
99.9% of attributes won't have a duplicate attribute.

This is more IL-level optimisation though 😅

TimLariviere (approved):
That's a good improvement :)

TimLariviere:
Would an `if` condition be better for the hot path than a `while` loop given 99.9% of the attributes won't be duplicated, though?
I guess the compiler already optimizes this kind of stuff.

e.g.
```fs
if scalars.[pos + 1].Key <> key then
    resultingIndex // Quick exit
else
    resultingIndex <- resultingIndex + 1

    while (...)
    
    resultingIndex
```

twop:
> Would an `if` condition be better for the hot path than a `while` loop given 99.9% of the attributes won't be duplicated? I guess the compiler already optimizes this kind of stuff.
> 
> e.g.
> 
> ```fsharp
> if scalars.[pos + 1].Key <> key then
>     pos // Quick exit
> else
>     resultingIndex <- resultingIndex + 1
> 
>     // Peek even further ahead
>     while (...)
>     
>     resultingIndex
> ```
It is possible, although we need to check out of bounds as well, regardless I couldn't measure any overhead with skipping (relative to what we have today, e.g. no overhead at all). So I think we should probably merge it as is and tinker about more performant implementation on a side "for fun" :)

PR #990, opened 2022-09-02
Maui Label changes

edgarfgp:
- Adds the new FontAutoScalingEnabledProperty See https://github.com/dotnet/maui/pull/1774
- Removes NamedSize from the font modifier as it is deprecated and will be deleted in NET 7 . See https://github.com/dotnet/maui/pull/5088


TimLariviere (approved):
Thanks a lot!
I removed the independent `fontAutoScalingEnabled` modifier because it's already settable through the `font` modifier

PR #995, opened 2022-09-03
Maui SearchBar changes

edgarfgp:
- Adds the new Properties
- Removes NamedSize from the font modifier as it is deprecated and will be deleted in NET 7 . See https://github.com/dotnet/maui/pull/5088


TimLariviere:
I'm noticing we let people set the cursor position and selection length but they can't listen for the user changing the cursor or the selection. This is a problem since Fabulous will override whatever the user is doing on each update.

Would be nice to make those modifiers both setter and event listener like the text/textChanged modifier.
```fs
static member inline cursorPosition(position: int, onPositionChanged: int -> 'msg)
```

PR #1004, opened 2022-09-05
Maui VisualElement  changes

edgarfgp:
- Adds missing MaximumHeightRequest, MaximumWidthRequest ,IsFocused
- Removed unsupported properties
- Window property is part of VisualElement but we are not using it in Fabulous.Maui.Controls . So it does not make sense to wrap it 
- ZIndex is part of VisualElement but it is internal so I can not add use in an attribute

### DIFF
--- a/samples/Maui/Playground/App.fs
+++ b/samples/Maui/Playground/App.fs
@@ -10,39 +10,58 @@ open Microsoft.Maui.Primitives
 open type Fabulous.Maui.View
 
 module App =
-    type Model = { HasNavigated: bool }
+    type FieldFocused =
+        | Entry1
+        | Entry2
+        | Entry3
+    
+    type Model = { Focus: FieldFocused option }
 
     type Msg =
-        | GoToA
-        | GoToB
+        | TextChanged of string
+        | FocusChanged of FieldFocused * bool
+        | SetFocus of FieldFocused option
 
-    let init () = { HasNavigated = false }
+    let init () = { Focus = None }
 
     let update msg model =
         match msg with
-        | GoToA -> { model with HasNavigated = false }
-        | GoToB -> { model with HasNavigated = true }
+        | TextChanged _ -> model
+        | FocusChanged (field, isFocused) ->
+            if isFocused then
+                { model with Focus = Some field }
+            else
+                { model with Focus = None }
+                
+        | SetFocus field -> { model with Focus = field }
+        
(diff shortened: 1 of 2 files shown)


TimLariviere:
Unfortunately `IsFocusedProperty` is read-only and will throw an exception if we try to set it.
I replaced it with a custom attribute calling `Focus()` / `Unfocus()` instead. Also merged the `IsFocus` field with its 2 events  `Focused` and `Unfocused` to make it easier to work with.

Here's an example for using it:
```fs
module App =
    type FieldFocused =
        | Entry1
        | Entry2
        | Entry3
    
    type Model = { Focus: FieldFocused option }

    type Msg =
        | TextChanged of string
        | FocusChanged of FieldFocused * bool
        | SetFocus of FieldFocused option

    let init () = { Focus = None }

    let update msg model =
        match msg with
        | TextChanged _ -> model
        | FocusChanged (field, isFocused) ->
            if isFocused then
                { model with Focus = Some field }
            else
                { model with Focus = None }
                
        | SetFocus field -> { model with Focus = field }
        
    let focusChanged field args =
        FocusChanged (field, args)

    let view model =
        Application(
            ContentPage(
                "Title",
                (VStack(spacing = 20.) {
                    let text = match model.Focus with None -> "None" | Some f -> f.ToString()
                    Label($"Field currently selected: {text}")
                    
                    Entry("Entry1", TextChanged)
                        .focus(model.Focus = Some Entry1, focusChanged Entry1)
                        
                    Entry("Entry2", TextChanged)
                        .focus(model.Focus = Some Entry2, focusChanged Entry2)
                        
                    Entry("Entry3", TextChanged)
                        .focus(model.Focus = Some Entry3, focusChanged Entry3)
                        
                    Button("Set focus on Entry1", SetFocus (Some Entry1))
                    Button("Set focus on Entry2", SetFocus (Some Entry2))
                    Button("Set focus on Entry3", SetFocus (Some Entry3))
                    Button("Unfocus", SetFocus None)
                })
                    .margin(20.)
            )
        )
```

<video src="https://user-images.githubusercontent.com/6429007/190602789-d52bc239-d912-4c4c-beaa-94e0b7ae1318.mp4" />

TimLariviere (approved):

edgarfgp:
Wow this looks perfect 👍 . We could do the same for v2 XF ?

PR #1008, opened 2022-09-08
fix(sidebar): refactor css rules to display sidebar on mobile devices

oscargonzalezdev:
For a better reading, I had to format the HTML and CSS files.
Only classes for the sidebar were updated.


oscargonzalezdev:
https://user-images.githubusercontent.com/86403175/189172994-6fd24771-1e27-4387-96f9-49bd7902dfff.mov

TimLariviere (changes requested):
@oscargonzalezdev That looks awesome! Thanks for working on this.

I'm sorry to tell you this now but the files you worked on are completely autogenerated and your changes will be lost immediately :x

Instead, you will need to get the content of the branch `docs`.
It contains all the template from which the website is generated (and the resulting generation is pushed to `gh-pages`).

Is it possible for you to apply your changes to the `docs` branch?
For example, the index file is located here: https://github.com/fsprojects/Fabulous/blob/docs/layouts/index.html

Just note that it's not pure HTML/CSS this time but a Hugo template (https://gohugo.io).

PR #1018, opened 2022-10-17
Use fully qualified namespace to avoid using Microsoft.Maui.IView

edgarfgp:
Fixes 

Screenshot 2022-10-17 at 17 08 51


TimLariviere (approved):
Awesome, thanks!

PR #1035, opened 2023-01-23
added ability to add attributes to the CollectionBuilder

twop:
- This is backward compatible change (purely additive)
- Added `AddScalar` to the `CollectionBuilder`
- Added `CollectionBuilder.yieldImpl` helper to not expose `MutStackArray1` as much, I think it will be easier to document
- We already had a usage in tests for such an API `Stack().automationId()`, thus no new tests were needed
- Note that now you have a choice either to add a property to `CollectionBuilder` or `WiddgetBuilder` or both. Meaning that we can control precisely at what point attributes can be declared

```fsharp
// this will add "(Stack() {...}).automationId("id)"
[<Extension>]
static member inline automationId<'msg, 'marker when 'marker :> IMarker>(this: WidgetBuilder<'msg, 'marker>, value: string) =
    this.AddScalar(Attributes.Automation.AutomationId.WithValue(value))

// this will add "Stack().automationId("id) {...}"
[<Extension>]
static member inline automationId<'msg, 'marker, 'itemMarker when 'marker :> IMarker>
    (
        this: CollectionBuilder<'msg, 'marker, 'itemMarker>,
        value: string
    ) =
    this.AddScalar(Attributes.Automation.AutomationId.WithValue(value))
```

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "fantomas": {
-      "version": "5.1.5",
+      "version": "5.2.0",
       "commands": [
         "fantomas"
       ]
--- a/src/Fabulous.Benchmarks/Benchmarks.fs
+++ b/src/Fabulous.Benchmarks/Benchmarks.fs
@@ -57,7 +57,9 @@ module DiffingAttributes =
 
     let update msg model =
         match msg with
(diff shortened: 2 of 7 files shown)


TimLariviere (approved):
That's a surprisingly easy change. Nice one 👍

TimLariviere:
Fantomas 5.2.0 doesn't have this issue. :)

PR #1067, opened 2024-02-11
Improve code documentation for MVU/Elm newbies

h0lg:
I'm currently wrapping my head around Elm, MVU and Fabulous, had a look at the changes in pre-release 2.5 #1065 6de97c06b0f2cc99a6349da8ec3177876953fe4a and tried to **update the code documentation** in a helpful way by connecting types to Elm or MVU concepts. This is an attempt to make Fabulous a bit more approachable for newbies like me who may or may not have done their homework learning about MVU and Elm.

**Please** **review** my changes **carefully** to make sure I'm not confusing things and **correct me** if I did!
You'll find some **questions** concerning opaque concepts like `CmdMsg` vs. `Cmd<'msg>` as well, which I mentioned in #927 .

### DIFF
--- a/src/Fabulous/Cmd.fs
+++ b/src/Fabulous/Cmd.fs
@@ -3,16 +3,35 @@
 open System.Threading
 open System.Threading.Tasks
 
-/// Dispatch - feed new message into the processing loop
+/// <summary>
+/// A function that feeds a new Message into the processing loop.
+/// <seealso href="https://elmish.github.io/elmish/#dispatch-loop" />
+/// </summary>
 type Dispatch<'msg> = 'msg -> unit
 
-/// Subscription - return immediately, but may schedule dispatch of a message at any time
+/// <summary>
+/// A function that returns immediately, but may schedule dispatch of one or multiple Messages at any time
+/// via its access to a Dispatch function.
+/// This is an abstraction over raw Messages passed along to the MVU processing loop
+/// necessary to deal with real-world scenarios in which Effect functions may take a while to complete.
+/// This models an Elm "side-effect" like in
+/// <seealso href="https://elmish.github.io/elmish/#tasks-and-side-effects" />
+/// or
+/// <seealso href="https://elmprogramming.com/side-effects.html" />
+/// </summary>
 type Effect<'msg> = Dispatch<'msg> -> unit
 
-/// Cmd - container for effects that may produce messages
+/// <summary>
+/// A list of Effects that may dispatch Messages;
+/// the carriers of instructions you issue from the init and update functions.
+/// See
+/// <seealso href="https://elmish.github.io/elmish/#commands" />
+/// </summary>
 type Cmd<'msg> = Effect<'msg> list
 
-/// Cmd module for creating and manipulating commands
+/// A module for creating and manipulating Commands
+/// with a Command being a list of Message-dispatching Effects you issue from the init and update functions
+/// and an Effect being a function with access to a Dispatch function receiving a Message.
 [<RequireQualifiedAccess>]
 module Cmd =
     /// Execute the commands using the supplied dispatcher
@@ -24,32 +43,48 @@ module Cmd =
             with ex ->
                 onError ex)
 
-    /// None - no commands, also known as `[]`
+    /// No command; an empty list of Message-dispatching Effects equivalent to `[]`.
+    /// For when you don't want to issue a Command.
     let none: Cmd<'msg> = []
 
-    /// When emitting the message, map to another type
+    /// <summary>
+    /// Converts a Command of type 'a into a Command of type 'msg.
+    /// This is useful for emitting Commands of a uniform type,
+    /// like when receiving child messages in a parent-child composition scenario. See
+    /// <seealso href="https://elmish.github.io/elmish/docs/parent-child.html" />
+    /// </summary>
     let map (f: 'a -> 'msg) (cmd: Cmd<'a>) : Cmd<'msg> =
         cmd |> List.map(fun g -> (fun dispatch -> f >> dispatch) >> g)
 
-    /// Aggregate multiple commands
+    /// <summary>
+    /// Concatenates the Effects of multiple Commands into one list.
+    /// Use for emitting multiple Commands at the same time from the init or update function.
+    /// E.g. in an parent-child composition scenario:
+    /// <seealso href="https://elmish.github.io/elmish/docs/parent-child.html" />
+    /// </summary>
     let batch (cmds: Cmd<'msg> list) : Cmd<'msg> = List.concat cmds
 
-    /// Command to call the effect
+    /// <summary>
+    /// Returns a command to call a custom Effect function with access to a Dispatch function.
+    /// Use for example to dispatch status updates or yield partial results from long-running background tasks.
+    /// </summary>
     let ofEffect (effect: Effect<'msg>) : Cmd<'msg> = [ effect ]
 
-    /// Command to issue a specific message
+    /// Command to issue a specific message.
+    /// Wraps the message into the Command structure returned from the update and init functions.
     let ofMsg (msg: 'msg) : Cmd<'msg> = [ fun dispatch -> dispatch msg ]
 
-    /// Command to issue a specific message, only when Option.IsSome = true
+    /// Command to issue the message from the message option if Option.IsSome
     let ofMsgOption (msg: 'msg option) : Cmd<'msg> =
         [ fun dispatch ->
               match msg with
               | None -> ()
               | Some msg -> dispatch msg ]
 
+    /// Creates Commands from the return values and/or exceptions of simple functions,
+    /// wrapping the call in a try/with statement. Use this to deal with code that may throw exceptions.
     module OfFunc =
-        /// Command to evaluate a simple function and map the result
-        /// into success or error (of exception)
+        /// Creates a Command to evaluate a simple function and map either the return value or exception to a message
         let either (task: 'a -> _) (arg: 'a) (ofSuccess: _ -> 'msg) (ofError: _ -> 'msg) : Cmd<'msg> =
             let bind dispatch =
                 try
@@ -59,7 +94,7 @@ module Cmd =
 
             [ bind ]
 
-        /// Command to evaluate a simple function and map the success to a message
+        /// Creates a Command to evaluate a simple function and map the return value to a message
         /// discarding any possible error
         let perform (task: 'a -> _) (arg: 'a) (ofSuccess: _ -> 'msg) : Cmd<'msg> =
             let bind dispatch =
@@ -70,7 +105,8 @@ module Cmd =
 
             [ bind ]
 
-        /// Command to evaluate a simple function and map the error (in case of exception)
+        /// Creates a Command to evaluate a simple function returning unit
+        /// and map the error (in case of exception) to a message
         let attempt (task: 'a -> unit) (arg: 'a) (ofError: _ -> 'msg) : Cmd<'msg> =
             let bind dispatch =
                 try
@@ -80,6 +116,9 @@ module Cmd =
 
             [ bind ]
 
+    /// Internal module for building Commands from the return values or exceptions of Async functions
+    /// using Async.Catch like an async try/with statement.
+    /// You'll probably want to use either of the modules OfAsync or OfAsyncImmediate instead of this.
     module OfAsyncWith =
         /// Command that will evaluate an async block and map the result
         /// into success or error (of exception)
@@ -98,6 +137,7 @@ module Cmd =
             [ bind >> start ]
 
         /// Command that will evaluate an async block and map the success
+        /// discarding any possible error
         let perform (start: Async<unit> -> unit) (task: 'a -> Async<_>) (arg: 'a) (ofSuccess: _ -> 'msg) : Cmd<'msg> =
             let bind dispatch =
                 async {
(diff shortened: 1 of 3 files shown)


src/Fabulous/Cmd.fs
```diff
@@ -80,6 +117,9 @@ module Cmd =
 
             [ bind ]
 
+    //TODO what do I use this for? Compared to other OfAsync... modules?
```
TimLariviere, line 120: `OfAsyncWith` is the base for `OfAsync` and `OfAsyncImmediate`. By itself, it lets you choose how you want the async computations to be executed. But in practice, you would use `OfAsync` when you want the async comp. to be executed on the thread pool, or `OfAsyncImmediate` when you want the async comp. to be executed on the current thread.

TimLariviere: Not much point to use `OfAsyncWIth` directly since the only 2 options are `Async.Start` and `Async.StartImmediate` which are covered by `OfAsync` and `OfAsyncImmediate`.

src/Fabulous/Cmd.fs
```diff
@@ -170,6 +214,8 @@ module Cmd =
         let inline attempt (task: 'a -> Async<_>) (arg: 'a) (ofError: _ -> 'msg) : Cmd<'msg> =
             OfAsyncWith.attempt Async.StartImmediate task arg ofError
 
+    //TODO what do I use this for? Compared to other OfAsync... modules?
```
TimLariviere, line 217: Most .NET libraries use `Task` rather than `Async` which is F#-only. `OfTask` is here when it's easier to just use `task` compared to `async`. Also, tasks are "hot" and will execute immediately as soon as they are defined compared to async which will wait to be executed (with `let!` or `Async.Start`, etc.).

src/Fabulous/Program.fs
```diff
@@ -3,12 +3,19 @@ namespace Fabulous
 open System
 open System.Diagnostics
 
+(*TODO Is either of these a program in the Elm sense? If so, where's the view in this one?
```
TimLariviere, line 6: Originally we had a single `Program` type containing the `view` function like in typical Elm. In Fabulous 3, I had to split the Program type in 2 types to enable support of MVU components because the view function is defined later.

```fs
let init = ...
let update = ...
let view = ...

let program =
    Program.stateful init update // Program<'arg, 'model, 'msg>
    |> Program.withView view // Program<'arg, 'model, 'msg, 'view>
``` 

```fs
let init = ...
let update = ...
let program = Program.stateful init update

let view =
    Component(program) {
        // the actual view function here
    }
```

TimLariviere: Do you think it would be better to rename the two Program types to avoid the confusion?

src/Fabulous/Program.fs
```diff
@@ -3,12 +3,19 @@ namespace Fabulous
...
+Or are these rather abstractions of or pre-cursors to an Elm Program?
+AFAIU in the Elm architecture a "program" manages the application's (or component's) state, actions, and view rendering.
+Please help me as a MVU/Elm newbie understand these types. *)
+//TODO what's the 'arg?
```
TimLariviere, line 10: `'arg` is a value you can pass to initialize a program. This `'arg` will be received by the `init` function.

```fs
module App =
    let init (isAdmin: bool) =
        { Model.Name = if isAdmin then "Admin" else "User" }
    
    let program = 
        Program.stateful init update
        |> Program.withView view

type MauiProgram =
    static member CreateMauiApp() =
        let isAdmin = true

        MauiApp
            .CreateBuilder()
            .UseFabulousApp(App.program, isAdmin)
```

src/Fabulous/Program.fs
```diff
@@ -19,10 +26,11 @@ type Program<'arg, 'model, 'msg> =
         ExceptionHandler: exn -> bool
     }
 
+//TODO how is this different to the above? What's a 'marker? what's the 'arg?
```
TimLariviere, line 29: `'marker` here is the type of the view. This is used by Fabulous.MauiControls and Fabulous.Avalonia to strongly-type the view being returned.

For example, a `ListView` only accepts "cells" (anything inheriting from `Cell` type like `TextCell`, `ImageCell` and so on). You don't want a page in a ListView. In Fabulous, this is represented by the marker interfaces: `IFabTextCell`, `IFabImageCell` which both inherit from `IFabCell`. `IFabPage` doesn't inherit from `IFabCell` making it impossible to be used inside a ListView.

TimLariviere: `'marker` in Program type is mainly used by components to signal that this component will return that specific view type and can only be used in a specific way.

We also use this `'marker` type when running an application. `UseFabulousApp` will require `'marker` to be `IFabApplication` because Maui/Avalonia requires an Application type and not some random Label for example.

src/Fabulous/Program.fs
```diff
@@ -58,23 +72,44 @@ module Program =
           Logger = ProgramDefaults.defaultLogger()
           ExceptionHandler = ProgramDefaults.defaultExceptionHandler }
 
+    //TODO when would I use this one? How does it compare to the other stateful* builders?
```
TimLariviere, line 75: `stateful` is the simplest type of program where a state is used. You send a message, the state is updated.

`statefulWithCmd` adds the concept of side effects on top of the state via `Cmd`s

`statefulWithCmdMsg` is the same than `statefulWithCmd` except it adds a layer of discriminated union to be easier to unit test. You can read more about it on https://zaid-ajaj.github.io/the-elmish-book/#/chapters/scaling/intent (Intent is just a different naming for the same concept)

TimLariviere: For `'arg`, please see explanation above

src/Fabulous/Program.fs
```diff
@@ -58,23 +72,44 @@ module Program =
...
+    /// See also
+    /// <seealso href="https://elmprogramming.com/elm-architecture-conclusion.html" />
+    /// ?
+    /// </summary>
     let statefulWithCmdMsg (init: 'arg -> 'model * 'cmdMsg list) (update: 'msg -> 'model -> 'model * 'cmdMsg list) (mapCmd: 'cmdMsg -> Cmd<'msg>) =
         let mapCmds cmdMsgs = cmdMsgs |> List.map mapCmd |> Cmd.batch
         define (fun arg -> let m, c = init arg in m, mapCmds c) (fun msg model -> let m, c = update msg model in m, mapCmds c)
 
+    (*TODO Subscriptions will be started or stopped automatically to match.
+        - I don't understand what that means. What (other?) Subscriptions - or - to match what?*)
```
TimLariviere, line 101: I think we are missing a word here. I guess it should be "... to match the lifecycle of the program".
Subscriptions are now disposed when the program exits, for example when a component is removed from the UI tree.

src/Fabulous/Program.fs
```diff
@@ -58,23 +72,44 @@ module Program =
...
     /// Subscribe to external source of events, overrides existing subscription.
     /// Return the subscriptions that should be active based on the current model.
     /// Subscriptions will be started or stopped automatically to match.
+    /// See also
+    /// <seealso href="https://elmprogramming.com/subscriptions.html" />
+    /// ?
+    /// </summary>
     let withSubscription (subscribe: 'model -> Sub<'msg>) (program: Program<'arg, 'model, 'msg>) = { program with Subscribe = subscribe }
 
+    //TODO In what scenario would I want to use this?
```
TimLariviere, line 112: Just like `Cmd.map`, `mapSubscription` converts child subscriptions to dispatch a parent type.

src/Fabulous/Cmd.fs
h0lg: Thanks for clarifying and especially for pointing out the difference in "hotness" - I understand F# better now :)
I'll update the docs with this info.

src/Fabulous/Program.fs
h0lg: @TimLariviere I've read about _Intent_ and understand that it's a pattern used in component programs for cleanly separating messages to the component from messages to the parent program. It's used to avoid leaking the decision into the parent program about which component messages to intercept, inspect or forward. The parent only needs to worry about handling the intents and can blindly pass component messages on to the component's update function.

The diff of https://github.com/fabulous-dev/Fabulous.Avalonia/pull/224, which switches the examples between between using `statefulWithCmdMsg` and `statefulWithCmd`, looks like what I've tried to describe [here](https://github.com/fabulous-dev/Fabulous.Avalonia/pull/224#issuecomment-1999759296).

I'm failing to see the connection between _Intent_ and _CmdMsg_ - other than there's command mapping and batching involved. Can you help me connect the dots?

TimLariviere: @h0lg True, sorry about the confusion. `Intent` is indeed another concept we called `ExternalMsg` in Fabulous to enable child -> parent communication. Not the same thing at all than `CmdMsg`.

Your understanding of `CmdMsg` is correct.

When side-effects are required, `init` and `update` can return a `Cmd<'msg>` along the updated `Model`. This `Cmd<'msg>` is essentially a list of functions to be executed by Fabulous as side-effects, which might dispatch new messages.

The nice thing with MVU is that it's very easy to unit test since the `init` and `update` functions are pure and will always return an updated state, so you can check that a specific message with a specific state will return the expected updated state.

But unit testing the returned `Cmd<'msg>` is extremely difficult as you would need to execute it yourself. What if that `Cmd<'msg>` is making a network call or database update?

It's easier to separate the "intention" (not `Intent`, that's where I got confused earlier) of side-effects with the actual implementation as it can just check against the "intention" (the `CmdMsg`) in the unit tests.

```fs
type Msg = | Clicked

let doNetworkCallCmd () =
    Cmd.OfAsync...

let update msg model =
    match msg with
    | Clicked -> { model with Clicked = true }, doNetworkCallCmd()

let ``When user clicks the button, do the network call``() =
    let initialState = { Clicked = false }
    let expectedState = { Clicked = true }

    let actualState, actualCmd = update Clicked initialState

    Assert.Equals(actualState, expectedState)
    // TODO: How do you test that "actualCmd" is doing the right thing without actually doing the network call?
```

```fs
type Msg = | Clicked

type CmdMsg = | DoNetworkCall

let doNetworkCallCmd () =
    Cmd.OfAsync...

let mapCmdMsgToMsg cmdMsg =
    match cmdMsg with
    | DoNetworkCall -> doNetworkCallCmd()

let update msg model =
    match msg with
    | Clicked -> { model with Clicked = true }, [ DoNetworkCall ]

let ``When user clicks the button, do the network call``() =
    let initialState = { Clicked = false }
    let expectedState = { Clicked = true }
    let expectedCmdMsg = DoNetworkCall

    let actualState, actualCmdMsg = update Clicked initialState

    Assert.Equals(actualState, expectedState)
    Assert.Equals(actualCmdMsg, expectedCmdMsg) // Way easier to unit test
```

h0lg: @TimLariviere I think I understand now - and thank you for helping me!

`statefulWithCmdMsg` allows you to refactor side-effects out of your `update` function, making it pure and easy to unit-test without executing the side-effects.

You do so by defining an extra discriminated union for messages that trigger side-effects (in your example called `CmdMsg`) and then return those messages from your update function instead of executing the side effects in place. 

An additional function (in your example called `mapCmdMsgToMsg`) maps each of those messages back to the side-effect.

If I understand correctly, I suggest that besides updating the code doco (which I'll try), your example would be a good addition to the doco as a topic called _Unit testing the update function_ or something similar.

To
- make this concept more accessible to newbies by getting closer to known ELM/MVU concepts,
- avoid confusion between `CmdMsg`, `Cmd<'msg>` and `ExternalMsg` and
- capture the variant of Program this creates,

I suggest the following renaming:

`CmdMsg` -> `SideEffect` (in examples - of course we can call it whatever we like)
`mapCmdMsgToMsg` -> `mapSideEffectToMsg` or simply `mapSideEffect`
`statefulWithCmdMsg` -> `statefulWithSideEffects` or `statefulWithIsolatedSideEffects`

h0lg: @TimLariviere I have updated the doco on the stateful* builders according to my new found understanding - thanks again for clarifying!

src/Fabulous/Program.fs
h0lg: @TimLariviere Isn't that what `Sub.map` is for?

src/Fabulous/Program.fs
h0lg: I've updated the doco changing the wording.

dsyme:
Closing this stale draft rather than merging. It contains unresolved review questions/TODOs, stale documentation URLs, and some guidance that needs revalidation against current async behavior. Documentation improvements should now target the consolidated docs and current Fabulous.Core API.

h0lg:
> Closing this stale draft rather than merging. It contains unresolved review questions/TODOs, stale documentation URLs, and some guidance that needs revalidation against current async behavior.

@dsyme This PR was an attempt to reach out to the maintainers to help me understand what I'm looking at in this sparsely documented source - and specifically how to relate concepts I find in here to what I read about ELM and MVU elsewhere.
It tried to get the ball rolling and got somewhere, but may have been too ambitious and stalled. Still, I think this source needs and deserves some helpful and accessible doco.

> Documentation improvements should now target the consolidated docs and current Fabulous.Core API.

I don't understand what that means. Are there new docs somewhere that I'm not aware of?
_Edit:_ Found them: https://fabulous-dev.github.io/Fabulous/docs/

MiroslavHustak:
@h0lg Please bear with us. Don Syme did a huge job during the last two days, but issues naturally come with a complete overhaul of the Fabulous repository. It is my task to address them; I started working on them today, and my list of known issues in the docs (such as broken links) is probably already longer than yours. Please follow the #fabulous project on Discord (fsharp).

h0lg:
@MiroslavHustak I understand - I've since found issues #1143 and #1156, which put things into perspective for me.

h0lg:
And thank you kindly @dsyme for the resurrection and @MiroslavHustak for picking up the maintenance <3
I had lost hope but love to see it happen! Just upgraded [SubTubular](https://github.com/h0lg/SubTubular) to Fabulous 10 and it seems to be running without a hitch. Looking forward to see more of Fabulous 🚀

PR #1074, opened 2024-03-22
Add support for Binding in Components

TimLariviere:
Components only supports local state through the use of `Context.State(defaultValue)`. 
With `Context.Binding(stateValue)`, a component can take a dependency on the local state of another component and be able to listen to changes as well as update the value, triggering a refresh of both components.

Usage:
```fs
// Children
let firstNameView (value: StateValue<string>) =
    Component() {
        let! firstName = Context.Binding(value)
        Entry(firstName.Current, firstName.Set)
    }

let lastNameView (value: StateValue<string>) =
    Component() {
        let! lastName = Context.Binding(value)
        Entry(lastName.Current, lastName.Set)
    }

// Parent
let parentView () =
    Component() {
        // Parent is the one owning the data
        let! firstName = Context.State("")
        let! lastName = Context.State("")

        VStack() {
            Label($"Full name is {firstName.Current} {lastName.Current}")
            firstNameView firstName
            lastNameView lastName
    }
```

### DIFF
--- a/src/Fabulous/Binding.fs
+++ b/src/Fabulous/Binding.fs
@@ -48,9 +48,11 @@ type BindingExtensions =
             | ValueSome d -> d.Dispose()
 
             // Subscribe to source context changes
+            let sourceKey = stateValue.Key
+
             let sub =
                 stateValue.Context.RenderNeeded.Subscribe(fun k ->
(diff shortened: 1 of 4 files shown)


edgarfgp (approved):
Awesome work

PR #1080, opened 2024-05-02
Guard against inner component null messages

edgarfgp:
In cases where you have a `Component()` which itself is in a `Component(program)` and a side effect happens in the inner component, the outer component will dispatch a message that is null.

So here we need make sure that we don't process null messages `ValueSome null`


edgarfgp:
I guess we no longer need this fix/check if we create dedicated non dispatch `define` functions.

PR #1081, opened 2024-05-13
Add new `CollectionBuilder` constructor using an `AttributesBundle`

edgarfgp:
Currently in Fabulous you can't pass a `WidgetBuilder` as part of the `CollectionBuilder` constructor using `AttributesBundle`.

### Before

Screenshot 2024-05-13 at 21 23 58

### After

Screenshot 2024-05-13 at 22 10 57

This PR changes the `CollectionBuilder` definition to use `AttributesBundle`.

### DIFF
--- a/src/Fabulous/Builders.fs
+++ b/src/Fabulous/Builders.fs
@@ -125,36 +125,45 @@ type Content<'msg> = { Widgets: MutStackArray1.T<Widget> }
 type CollectionBuilder<'msg, 'marker, 'itemMarker> =
     struct
         val WidgetKey: WidgetKey
-        val Scalars: StackList<ScalarAttribute>
         val Attr: WidgetCollectionAttributeDefinition
+        val Scalars: AttributesBundle
 
-        new(widgetKey: WidgetKey, scalars: StackList<ScalarAttribute>, attr: WidgetCollectionAttributeDefinition) =
+        new(widgetKey: WidgetKey, attr: WidgetCollectionAttributeDefinition, scalars: AttributesBundle) =
             { WidgetKey = widgetKey
               Scalars = scalars
               Attr = attr }
 
         new(widgetKey: WidgetKey, attr: WidgetCollectionAttributeDefinition) =
             { WidgetKey = widgetKey
-              Scalars = StackList.empty()
+              Scalars = AttributesBundle(StackList.empty(), ValueNone, ValueNone)
               Attr = attr }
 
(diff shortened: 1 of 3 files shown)


src/Fabulous/Builders.fs
```diff
@@ -125,36 +125,45 @@ type Content<'msg> = { Widgets: MutStackArray1.T<Widget> }
 type CollectionBuilder<'msg, 'marker, 'itemMarker> =
     struct
         val WidgetKey: WidgetKey
-        val Scalars: StackList<ScalarAttribute>
         val Attr: WidgetCollectionAttributeDefinition
+        val Scalars: AttributesBundle
```
TimLariviere, line 129:
```suggestion
        val Attributes: AttributesBundle
```

src/Fabulous/Builders.fs
```diff
@@ -125,36 +125,45 @@ type Content<'msg> = { Widgets: MutStackArray1.T<Widget> }
...
 
-        new(widgetKey: WidgetKey, scalars: StackList<ScalarAttribute>, attr: WidgetCollectionAttributeDefinition) =
+        new(widgetKey: WidgetKey, attr: WidgetCollectionAttributeDefinition, scalars: AttributesBundle) =
             { WidgetKey = widgetKey
               Scalars = scalars
               Attr = attr }
```
TimLariviere, line 134: Good move, it is best to keep the same ordering of the parameters. Just watch out for the correct naming.
Scalars != Attributes

Also, it would be best to keep the original constructor as a short-hand for existing code.

```suggestion
        new(widgetKey: WidgetKey, attr: WidgetCollectionAttributeDefinition, attributes: AttributesBundle) =
            { WidgetKey = widgetKey
              Attr = attr
              Attributes = attributes }
              
        new(widgetKey: WidgetKey, attr: WidgetCollectionAttributeDefinition, scalars: StackList<ScalarAttribute>) =
            { WidgetKey = widgetKey
              Attr = attr
              Attributes = AttributesBundle(scalars, ValueNone, ValueNone) }
```

TimLariviere (changes requested):
Good addition, This would be useful for more advanced scenario when dealing with CollectionBuilder. I made some comments.
Mainly watch out for the difference between Scalar and AttributeBundle.

src/Fabulous/Builders.fs
TimLariviere: Technically, we could release this change to the current v2 release, but we would need to keep the parameters ordering

edgarfgp: Yes would be good to release in 2.x also

PR #1084, opened 2024-10-24
Add equality constraint on 'msg to prevent confusion with new non-MVU modifiers

TimLariviere:
This pull request adds an equality constraint on `'msg` on most types to prevent confusions with the new modifiers for non-MVU components.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,10 +3,11 @@
   "isRoot": true,
   "tools": {
     "fantomas": {
-      "version": "6.2.3",
+      "version": "6.3.15",
       "commands": [
         "fantomas"
-      ]
+      ],
+      "rollForward": false
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 _No unreleased changes_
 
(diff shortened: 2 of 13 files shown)


edgarfgp:
Thanks Tim :)

kerams:
Is the constraint just artificial or are messages actually compared to each other? My main message type has `NoEquality`, so I'd like to know what the ramifications of doing this instead would be

```
[<NoComparison; CustomEquality>]
type Msg =
    ...

    interface IEquatable<Msg> with
        member this.Equals (other: Msg | null) = Object.ReferenceEquals this other

**Comment by TimLariviere** [comment]
It is indeed artificial. We don't compare messages.

This constraint on equality is to trick the F# compiler to differentiate between `'msg` and function types (eg. `unit -> unit`) as functions are never equatable.

This is useful to support both implicit dispatch and non-MVU widgets at the same time.

```fs
// without equality constraint,
// this is wrongly assumed as
// 'msg = unit -> unit
// Fabulous will try to dispatch that function
// instead of simply calling it

Button("Click me", fun() -> printfn "hello")
```

TimLariviere:
Do you have a specific requirement to not have your messages automatically be equatable?

As a workaround, implementing IEquatable by yourself should be fine.

kerams:
Thanks for the response. My messages contain data that does not have equality, and I am also actively annotating messages types with `NoComparison; NoEquality` in order not to generate unneeded members.

PR #1088, opened 2024-11-16
Remove System.Memory

kerams:
I don't think this is needed.


TimLariviere:
Thanks! That's indeed not being used.

PR #1091, opened 2024-11-24
Refactor EnumerationMode

kerams:
[`struct` adds extra overhead.](https://sharplab.io/#v2:DYLgZgzgPg9gDgUwHYAIDKBPCAXBBbAOgCUBXJbASzwQIGEY84LgEAnNNgNwoGMEIAsAChhAbQA8abKxI9sAbhQBJCACEMRBGAAyFANYJFmgI4kKrBAEUSAQ2AUwFBABMAgjz4QIigHIx6jDasFBAwSL4wAKKmdhTYGAB8ALrC8YgokUgk1Kw2lGEAsjDOCOIA5DYJKAC8wij1KFAorsDArs4lzgDyrJp4MJwuKDBgKDgycigAFBWiSSgAVCgARjAwwACUdQ1iktKyCspqGlq6BkYIpuZWtvaOLu6e3ih+AXBBIWER0bdxiSlCNIIDJZHJ5CiFYoIABM5UqNW29SaLTaHRcPT6AyGIxQs3mS1W6yAA==) because case elements are not stored in the tupled form but as separate fields in the struct.

### DIFF
--- a/src/Fabulous/WidgetDiff.fs
+++ b/src/Fabulous/WidgetDiff.fs
@@ -11,9 +11,9 @@ type ScalarAttributeComparison =
 
 [<Struct; IsByRefLike; RequireQualifiedAccess; NoComparison; NoEquality>]
 type EnumerationMode<'a> =
-    | AllAddedOrRemoved of struct ('a[] * bool)
+    | AllAddedOrRemoved of prev: 'a[] * bool // the first element is either prev or next depending on the bool, but using prev as the label allows the compiler to reuse struct fields between cases
     | Empty
-    | ActualDiff of prevNext: struct ('a[] * 'a[])
+    | ActualDiff of prev: 'a[] * next: 'a[]
 
 module EnumerationMode =
     let fromOptions prev next =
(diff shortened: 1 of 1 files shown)


kerams:
We could go further with

```fsharp
[<Struct; IsByRefLike; RequireQualifiedAccess; NoComparison; NoEquality>]
type EnumerationMode<'a> =
    | AllAdded of prev: 'a[] 
    | AllRemoved of prev: 'a[] 
    | Empty
    | ActualDiff of prev: 'a[] * next: 'a[]
```

This would get rid of the bool in the DU struct at the cost of slightly confusing case field names. (Althought struct field merging might require F# 9 and SDK 9)

kerams:
ping

edgarfgp:
> This would get rid of the bool in the DU struct at the cost of slightly confusing case field names. (Although struct field merging might require F# 9 and SDK 9)

We can not update the target to NET 9 because `Fabulouls.Avalonia` still targets NET 8.

edgarfgp (approved):
Thanks Kerams

PR #1095, opened 2024-12-15
Remove unnecessary ValueOptions

kerams:
This decreases the size of `AttributesBundle` and `Widget` structs by 24 and 32 bytes, respectively.

There's no downside of using an empty array instead of `ValueNone`, since `[||]` is lowered to `System.Array.Empty<_>`, which doesn't allocate either. The code also didn't seem to rely on the distinction between `ValueNone` and `ValueSome [||]`.


src/Fabulous/Builders.fs
```diff
@@ -7,7 +7,7 @@ open Fabulous.StackAllocatedCollections
 open Fabulous.StackAllocatedCollections.StackList
 open Microsoft.FSharp.Core
 
-type AttributesBundle = (struct (StackList<ScalarAttribute> * WidgetAttribute[] voption * WidgetCollectionAttribute[] voption * EnvironmentAttribute[] voption))
```
edgarfgp, line 10: I like that we can remove `ValueSome and ValueNone` when using it. Only downside is that we will need to update the `Avalonia and MauiControls` in many places :(

TimLariviere:
Thanks for the PR. That's would be a huge improvement given how many AttributesBundle are created all the time!
I will play with it this week and merge it :)

src/Fabulous/Array.fs
```diff
@@ -61,7 +61,7 @@ module Array =
     let inline sortInPlace<'T, 'V when 'V: comparison> ([<InlineIfLambda>] getKey: 'T -> 'V) (attrs: 'T[]) : 'T[] =
         let N = attrs.GetLength(0)
 
-        for i in [ 1 .. N - 1 ] do
+        for i in 1 .. N - 1 do
```
kerams, line 64: This is very important -
https://sharplab.io/#v2:DYLgZgzgPsCmAuACAHogXogvAWAFCIMTAHsAnRAS0oDtEBGRAOkfoAZEATYvQ39SnnwJ44SAJ78c+QiXJUKtANr0mLOuwC6nbtKEYKQA

TimLariviere:
Merged, thanks!
Gonna release it quickly

src/Fabulous/Builders.fs
kerams: @edgarfgp, happy to take care of the Avalonia part.

edgarfgp: Thanks @Kerams. I will appreciate it.

TimLariviere: Ah, I've actually submitted the PR already 😅 
https://github.com/fabulous-dev/Fabulous.Avalonia/pull/277

PR #1098, opened 2025-01-18
Add YieldFrom to CollectionBuilder

kerams:


src/Fabulous/Builders.fs
```diff
@@ -241,6 +241,14 @@ type CollectionBuilder<'msg, 'marker, 'itemMarker when 'msg: equality> =
 
             res
 
+        member inline _.YieldFrom(sequence: WidgetBuilder<'msg, 'itemMarker> seq) : Content<'msg> =
+            // TODO optimize this one with addMut
```
kerams, line 245: I'm not really sure `addMut` would be more efficient than this.

PR #1101, opened 2025-02-28
target net8 to fix attribute targets incompatibilities.

edgarfgp:
In NET 9 the F# compiler has a better analysis  for `AttributeTargets` but this means that we will need to use NET 8 minimum and FSharp.Core `8.0.300` to be able to build with NET 9 installed.


TimLariviere (approved):

TimLariviere:
Given .NET 7 has reach End of Life in May 2024 and .NET 8 is LTS, we are safe to move to .NET 8

PR #1103, opened 2025-05-12
Optimize

kerams:


src/Fabulous/Array.fs
```diff
@@ -89,7 +89,7 @@ module StackAllocatedCollections =
         [<NoComparison; NoEquality>]
         type private Part<'v> =
             | Empty
-            | Filled of struct (Items<'v> * Part<'v>)
+            | Filled of Items<'v> * Part<'v>
```
kerams, line 92: Removes indirection.

src/Fabulous/Array.fs
```diff
@@ -258,8 +258,8 @@ module StackAllocatedCollections =
 
     [<Struct; NoComparison; NoEquality>]
     type StackArray3<'v> =
-        | Few of data: struct (Size * 'v * 'v * 'v)
         | Many of arr: 'v array
+        | Few of data: Size * 'v * 'v * 'v
```
kerams, line 262: Removes indirection and decreases total DU struct size with some type parameters.

src/Fabulous/Array.fs
```diff
@@ -609,10 +607,9 @@ module StackAllocatedCollections =
 
         [<Struct>]
         type Op =
-            | Added of added: uint16
-            | Removed of removed: uint16
-            | Changed of changed: uint16
-
+            | Added of value: uint16
+            | Removed of value: uint16
+            | Changed of value: uint16
```
kerams, line 612: Merges fields and reduces struct size from 12 to 8 bytes.

edgarfgp:
Thanks @kerams

PR #1104, opened 2025-05-12
Use task CE for task commands

kerams:


src/Fabulous/Cmd.fs
```diff
@@ -172,21 +172,60 @@ module Cmd =
...
+                            do! task arg
+                        with e ->
+                            ofError e |> dispatch
+                    }
+                    |> ignore<Task<_>>
+            ]
+
+        let inline msg (task: Task<'msg>) = perform (fun () -> task) () id
 
         let inline msgOption (task: Task<'msg option>) =
```
kerams, line 217: Tasks are hot, so these 2 should probably be `task: unit -> Task<_>`, but I've left it as it is.

edgarfgp (commented):
Thanks

PR #1105, opened 2025-06-05
Ensure old widgets are disposed when replaced by new ones

TimLariviere:
I noticed that subscriptions from a component were not being disposed when the component was replaced. After investigating, I realized we were not properly disposing removed widgets.

In particular:
- When replacing one widget with another, the old widget's `ViewNode` was never disposed — the new widget was injected into the tree, but the old one remained in memory.
- When removing a widget from its parent without replacement, it also wasn't disposed.
- The same applied to the removal of entire WidgetCollections (e.g., when GestureRecognizers were present in the previous state but removed in the new one).

This PR fixes the issue by ensuring that dispose is correctly called on the ViewNode of any removed widget, enabling proper cleanup and avoiding potential memory/resource leaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 _No unreleased changes_
 
+## [3.0.0-pre23] - 2025-06-05
+
+### Changed
+- Ensure old widgets are disposed when replaced by new ones by @TimLariviere in https://github.com/fabulous-dev/Fabulous/pull/1105
+
 ## [3.0.0-pre22] - 2025-05-18
 
 ### Changed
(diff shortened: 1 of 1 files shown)


edgarfgp (approved):

src/Fabulous/Attributes.fs
```diff
@@ -255,8 +262,21 @@ module Attributes =
...
-        let updateNode _ (newValueOpt: ArraySlice<Widget> voption) (node: IViewNode) =
+        let updateNode (oldValueOpt: ArraySlice<Widget> voption) (newValueOpt: ArraySlice<Widget> voption) (node: IViewNode) =
             let targetColl = getCollection node.Target
+
+            if oldValueOpt.IsSome then
+                // Dispose the existing children if they exist
+                let oldWidgets = oldValueOpt.Value
+                let span = ArraySlice.toSpan oldWidgets
+
+                for index = 0 to span.Length do
```
kerams, line 273: Isn't this off by one?

src/Fabulous/WidgetDiff.fs
```diff
@@ -83,7 +83,7 @@ and [<Struct; RequireQualifiedAccess>] WidgetChange =
     | Added of widget: WidgetAttribute
     | Removed of widget: WidgetAttribute
     | Updated of widget: WidgetAttribute * diff: WidgetDiff // updated * diff
-    | ReplacedBy of widget: WidgetAttribute
+    | ReplacedBy of oldWidget: WidgetAttribute * newWidget: WidgetAttribute
```
kerams, line 86: Using these labels will almost double the struct size. If you reuse `widget` and `diff`, it will remain the same. (I understand that the labels might not be very descriptive with respect to the DU case, which is why I used that comment above. It's annoying, but worth it in my opinion.)

PR #1106, opened 2025-06-05
Mark non MVU components as experimental

edgarfgp:
As we work and refine more the `State, Binding` Environment and EnvironmentObject` and to make sure people expectations are clear. I propose we mark these as Experimental.


dsyme:
Closing rather than merging. This conflicts with current main, removes extension-type attributes, and introduces experimental warnings without an agreed project-wide experimental API policy. These component APIs have since shipped and should be evolved through focused API proposals.

PR #1108, opened 2025-11-12
feat: add onReceive modifier

edgarfgp:


dsyme:
Integrated and hardened for the consolidated layout in commit 59defe23d56b69a02baeb5d58361d54e4cbfc1ea. The port uses deterministic synchronous event tests and retains original-author attribution.

PR #1321, opened 2026-09-15
Update Avalonia package versions to 12.1.1 in templates

kant2002:


MiroslavHustak:
Currently taking care (Sep 17, 2026)

